PR #1, opened 2022-01-06
chore(action): add new assert-camunda-git-emails checker

cmur2:
Related to INFRA-2954

### DIFF
--- a/assert-camunda-git-emails/action.yml
+++ b/assert-camunda-git-emails/action.yml
@@ -1,0 +1,22 @@
+---
+name: Assert Camunda Git Emails
+
+description: Asserts that only Camunda email addresses are used in Git commits.
+
+inputs: {}
+
+runs:
+  using: composite
+  steps:
+  - uses: actions/checkout@v2
+    if: ${{ github.event.pull_request }}
+    with:
+      fetch-depth: 0
+      ref: ${{ github.event.pull_request.head.sha }}
+
+  - name: Check for violating email addresses
+    if: ${{ github.event.pull_request }}
+    shell: bash
+    env:
+      GIT_RANGE: "origin/${{ github.base_ref }}...${{ github.event.pull_request.head.sha }}"
+    run: ${{ github.action_path }}/check.sh
--- a/assert-camunda-git-emails/action.yml
+++ b/assert-camunda-git-emails/action.yml
@@ -3,7 +3,10 @@ name: Assert Camunda Git Emails
 
 description: Asserts that only Camunda email addresses are used in Git commits.
 
-inputs: {}
+inputs:
+  additional-allowed-emails-regex:
+    description: 'String containing a regex expression for the grep tool which can contain more emails to allow.'
+    default: ''
 
 runs:
   using: composite
@@ -19,4 +22,5 @@ runs:
     shell: bash
     env:
       GIT_RANGE: "origin/${{ github.base_ref }}...${{ github.event.pull_request.head.sha }}"
+      ADDITIONAL_ALLOWED_EMAILS_REGEX: "${{ inputs.additional-allowed-emails-regex }}"
     run: ${{ github.action_path }}/check.sh
--- a/assert-camunda-git-emails/check.sh
+++ b/assert-camunda-git-emails/check.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -eu
+
+echo "Looking at Git commits in range $GIT_RANGE"
+
+echo "All found email addresses:"
+git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | sort | uniq
+
+# shellcheck disable=SC2126
+NUM_VIOLATIONS=$(git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | \
+  grep -v "@camunda.com\|@users.noreply.github.com\|noreply@github.com" | \
+  wc -l)
+
+if [ "$NUM_VIOLATIONS" -ne 0 ]; then
+  echo "Failure! Found $NUM_VIOLATIONS violating email addresses:"
+  git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | \
+    grep -v "@camunda.com\|@users.noreply.github.com\|noreply@github.com" | \
+    sort | uniq
+
+  exit 1
+fi
+
+exit 0
--- a/assert-camunda-git-emails/check.sh
+++ b/assert-camunda-git-emails/check.sh
@@ -2,6 +2,15 @@
 
 set -eu
 
+# This variable has to contain regex expression for the grep tool
+HARDCODED_ALLOWED_EMAILS_REGEX="@camunda.com\|@users.noreply.github.com\|noreply@github.com"
+
+if [ "$ADDITIONAL_ALLOWED_EMAILS_REGEX" = "" ]; then
+  ALLOWED_EMAILS_REGEX="$HARDCODED_ALLOWED_EMAILS_REGEX"
+else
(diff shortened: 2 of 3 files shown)


assert-camunda-git-emails/check.sh
```diff
@@ -0,0 +1,28 @@
...
+set -eu
+
+echo "Looking at Git commits in range $GIT_RANGE"
+
+echo "All found email addresses:"
+git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | sort | uniq
+
+# shellcheck disable=SC2126
+NUM_VIOLATIONS=$(git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | \
+  grep -v "@camunda.com" | \
```
maxdanilov, line 12: can we make this whitelist configurable?
e.g. if I want to allow "something@renovate.com" without forking the action.

maxdanilov: camunda.com et al. can stay hardcoded, but the extensibility would be nice to have.

assert-camunda-git-emails/check.sh
```diff
@@ -0,0 +1,28 @@
...
+  grep -v "@users.noreply.github.com" | \
+  grep -v "noreply@github.com" | \
+  wc -l)
+
+if [ "$NUM_VIOLATIONS" -ne 0 ]; then
+  echo "Failure! Found $NUM_VIOLATIONS violating email addresses:"
+  git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | \
+    grep -v "@camunda.com" | \
```
maxdanilov, line 20: You can use a regex with multiple values in a list:
```suggestion
    grep -v "@camunda.com\|@users.noreply.github.com\|noreply@github.com" 
```

assert-camunda-git-emails/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  steps:
+  - uses: actions/checkout@v2
+    with:
+      fetch-depth: 0
+      ref: ${{ github.event.pull_request.head.sha }}
+
+  - name: Check for violating email addresses
+    shell: bash
+    env:
+      GIT_RANGE: "origin/${{ github.base_ref }}...${{ github.event.pull_request.head.sha }}"
```
maxdanilov, line 19: would it make sense to do safety check in the action and do nothing if the action is run outside of the PR context?
it case `on: [pull_request]` is omitted.

maxdanilov (changes requested):
Pretty nice implementaition, left some comments.

Also, what versioning model we'll have for the action? Semver?

assert-camunda-git-emails/check.sh
cmur2: > without forking the action
I imagined this action to be the source of truth and people to raise a PR if they want to add an exclusion. That said, I can make it configurable via an environment variable :+1:

assert-camunda-git-emails/check.sh
cmur2: Applied 🙇🏼

assert-camunda-git-emails/check.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -eu
+
+echo "Looking at Git commits in range $GIT_RANGE"
+
+echo "All found email addresses:"
+git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | sort | uniq
+
+# shellcheck disable=SC2126
+NUM_VIOLATIONS=$(git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | \
```
maxdanilov, line 11: what do you think about
```suggestions
VIOLATIONS=$(git log --pretty=tformat:"%ae%n%ce" "$GIT_RANGE" | \
grep -v "@camunda.com\|@users.noreply.github.com\|noreply@github.com")
NUM_VIOLATIONS=$(echo $VIOLATIONS | wc -l)
UNIQUE_VIOLATIONS=$(echo $VIOLATIONS | sort | uniq)
```
and use those later in the logic, to avoid code duplication?

cmur2: I tried it just now but it gets ugly quickly e.g. when trying to count the number of lines stored in a bash variable (`echo "" | wc -l` yields 1 and `echo "foo" | wc -l` also yields 1).

In my initial proposal I tried to keep this bash script as dumb as possible to avoid nasty edge cases. If it grows more complex, we should migrate to Python. I think the code duplication avoided is not that big (and part will be achieved with having an environment variable with a default list of email addresses).

assert-camunda-git-emails/check.sh
cmur2: Please have a look at the implementation in https://github.com/camunda/infra-global-github-actions/pull/1/commits/57344c0251a4444598ffa30b0e67edd76b36666b (lacks docs yet). This requires the user to provide an expression suitable for `grep -v` directly (thus requiring `\|` separators between OR-ed parts) instead of some generic e.g. comma-separated list. The reason here is that I want to keep the complexity of this script to a minimum due to using Bash - everything requiring string manipulations I think we need to switch to something like Python.

Looking at the implementation I'm not sure if the extensibility added is worth it given the applied limitations, though. WDYT?

cmur2:
> Also, what versioning model we'll have for the action? Semver?

I outlined in the README of this action that `- uses: camunda/infra-global-github-actions/assert-camunda-git-emails@main` should be used effectively depending on the main branch. Reasoning here is that I think it's too early to come up with a versioning model when we just have one action. In the future I'd strive to adopt something like what Github does with `v1`, `v2`, etc. branches/moving tags for major changes but we'd have to consider first how this work with together with other actions in this repo.

assert-camunda-git-emails/check.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -eu
+
+# This variable has to contain a `grep` expression
+HARDCODED_ALLOWED_EMAILS_LIST="@camunda.com\|@users.noreply.github.com\|noreply@github.com"
```
maxdanilov, line 6: maybe add the regex nature to the naming? e.g.
```suggestion
HARDCODED_ALLOWED_EMAILS_REGEX="@camunda.com\|@users.noreply.github.com\|noreply@github.com"
```

assert-camunda-git-emails/check.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ "$ADDITIONAL_ALLOWED_EMAILS_LIST" = "" ]; then
+  ALLOWED_EMAILS_LIST="$HARDCODED_ALLOWED_EMAILS_LIST"
```
maxdanilov, line 9: same
```suggestion
  ALLOWED_EMAILS_REGEX="$HARDCODED_ALLOWED_EMAILS_LIST\|$ADDITIONAL_ALLOWED_EMAILS_LIST"
```

assert-camunda-git-emails/check.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -eu
+
+# This variable has to contain a `grep` expression
+HARDCODED_ALLOWED_EMAILS_LIST="@camunda.com\|@users.noreply.github.com\|noreply@github.com"
+
+if [ "$ADDITIONAL_ALLOWED_EMAILS_LIST" = "" ]; then
```
maxdanilov, line 8: same
```suggestion
if [ "$ADDITIONAL_ALLOWED_EMAILS_REGEX" = "" ]; then
```

maxdanilov (commented):
Looks good, some small comments from my side.

assert-camunda-git-emails/check.sh
cmur2: Good point!

maxdanilov (approved):
Looks great 🚀

PR #16, opened 2022-10-25
add preview-env destroy action

chaima-mnsr:
Related to: https://github.com/camunda/team-infrastructure/issues/102

### DIFF
--- a/preview-env/destroy/action.yml
+++ b/preview-env/destroy/action.yml
@@ -1,0 +1,64 @@
+---
+name: Destroy preview environment
+
+description: Destroy preview environment.
+
+inputs:
+  app_name:
+    description: Argocd app name to be deleted.
+    required: true
+
+  revision:
+    description: Revision or branch of the Helm chart. Typically the branch to be deployed, or `master`.
+    required: true
+
+  argocd_server:
+    description: URL of the Argo CD instance to target.
+    required: false
+    default: argocd.int.camunda.com
+
+  argocd_token:
+    description: An Argo CD token with sufficient permissions to create Applications.
+    required: true
+
+  argocd_version:
+    default: v2.3.1
+    description: Version tag of Argo CD CLI tool
+    required: false
+runs:
+  using: composite
+  steps:
+  - name: Restore cache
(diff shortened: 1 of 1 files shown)


preview-env/destroy/action.yml
```diff
@@ -0,0 +1,95 @@
...
+      if [ "${{ inputs.app_name }}" == "" ]; then
+        APP_NAME="$(argocd app list -o json | jq -r '.[] | select(.spec.source.targetRevision=="${{ inputs.deployment_ref }}").metadata.name')"
+      else
+        APP_NAME="${{ inputs.app_name }}"
+      fi
+      echo "APP_NAME=${APP_NAME}" >> ${GITHUB_ENV}
+    env:
+      ARGOCD_AUTH_TOKEN: ${{ inputs.argocd_token }}
+      ARGOCD_OPTS: "--grpc-web"
+      ARGOCD_SERVER: ${{ inputs.argocd_server  }}
```
Langleu, line 68: Here I've already wondered in the previous per team action why we don't just supply the `app_name`. We know in 100% of the cases what the app is called, so why "overcomplicate" it by trying to find it and parse it.

preview-env/destroy/action.yml
```diff
@@ -0,0 +1,95 @@
...
+  - name: Wait for sync (up to 20 minutes)
+    shell: bash
+    run: |-
+      echo "::group::wait for sync"
+      argocd app wait ${{ env.APP_NAME }} --timeout 1200 --operation
+      echo "::endgroup::"
+    env:
+      ARGOCD_AUTH_TOKEN: ${{ inputs.argocd_token }}
+      ARGOCD_OPTS: "--grpc-web"
+      ARGOCD_SERVER: ${{ inputs.argocd_server  }}
```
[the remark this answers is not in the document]
chaima-mnsr: Make sense, not really needed in the teardown part, I will remove it.

preview-env/destroy/action.yml
chaima-mnsr: That is right 💯 , we can simplify this.

PR #25, opened 2023-01-25
chore(gha) Adding optional 'build-args' input to docker build step.

nwhinkle:
This PR will add docker _build-args_ as an optional input for the docker _build_ action.

Related: camunda/team-infrastructure#144

### DIFF
--- a/build-docker-image/action.yml
+++ b/build-docker-image/action.yml
@@ -16,6 +16,9 @@ inputs:
   image_name:
     description: Docker image name WITHOUT registry and WITHOUT Docker tag, e.g. example/image.
     required: true
+  build_args:
+    description: Docker build args to append to `docker build` command.
+    required: false
 
 runs:
   using: composite
@@ -27,6 +30,7 @@ runs:
       echo "-----"
       echo "Registry host: ${{ inputs.registry_host }}"
       echo "Image name:    ${{ inputs.image_name }}"
+      echo "Build-Args:    ${{ inputs.build_args }}"
 
   - name: Validate inputs
     shell: bash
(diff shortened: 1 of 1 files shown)


build-docker-image/action.yml
```diff
@@ -79,6 +82,7 @@ runs:
     uses: docker/build-push-action@v3
     with:
       context: .
+      build-args: ${{ inputs.build_args }}
```
cmur2, line 85: Since `inputs.build_args` is optional did you check what happens if it is not set? Does everything still work and it is assumed to be empty string orso

cmur2 (commented):
One question before approving this

build-docker-image/action.yml
nwhinkle: Good question. It wil skip over/ignore it if it's not set. Confirmed in this [build](https://github.com/camunda/camunda-docker-ci-chrome/actions/runs/4005644476/jobs/6876251749) by setting it to build against this branch but without passing the _build-args_ to the action.

build-docker-image/action.yml
```diff
@@ -16,6 +16,9 @@ inputs:
   image_name:
     description: Docker image name WITHOUT registry and WITHOUT Docker tag, e.g. example/image.
     required: true
+  build_args:
+    description: Docker build args to append to `docker build` command.
+    required: false
 
 runs:
   using: composite
```
cmur2, line 24: For debugging purposes it would also be nice to show the new input with `echo` in the `"Echo inputs"` step please

nwhinkle: Added this.

cmur2 (approved):

PR #26, opened 2023-01-25
Add action for using CI Nexus cache

chaima-mnsr:
Related to: https://github.com/camunda/team-infrastructure/issues/282

### DIFF
--- a/use-ci-nexus-cache/action.yml
+++ b/use-ci-nexus-cache/action.yml
@@ -1,0 +1,42 @@
+---
+name: Use CI Nexus cache
+
+description: Sets up use of co-located pull-through cache for Maven artifacts via ~/.m2/settings.xml.
+
+inputs:
+  server_username:
+    description: 'Nexus Username'
+  server_password:
+    description: 'Nexus Password'
+  server_id:
+    description: 'server ID'
+
+runs:
+  using: composite
+  steps:
+  - name: Create ~/.m2/settings.xml with cache configuration
+    shell: bash
+    run: |
+      cat <<EOF > ~/.m2/settings.xml
+      <?xml version="1.0" encoding="UTF-8"?>
+      <settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
+        <interactiveMode>false</interactiveMode>
+        <servers>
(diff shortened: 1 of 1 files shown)


use-ci-nexus-cache/action.yml
```diff
@@ -0,0 +1,42 @@
...
+    run: |
+      cat <<EOF > ~/.m2/settings.xml
+      <?xml version="1.0" encoding="UTF-8"?>
+      <settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
+        <interactiveMode>false</interactiveMode>
+        <servers>
+          <server>
+            <id>${{ inputs.server_id }}</id>
```
maxdanilov, line 28: what is `server_id`? does it have to be dynamic?

use-ci-nexus-cache/action.yml
```diff
@@ -0,0 +1,42 @@
...
+            <username>${{ inputs.server_username }}</username>
+            <password>${{ inputs.server_password }}</password>
+          </server>
+        </servers>
+        <mirrors>
+          <mirror>
+            <id>${{ inputs.server_id }}</id>
+            <mirrorOf>*</mirrorOf>
+            <name>Camunda Nexus</name>
+            <url>https://repository.nexus.camunda.cloud/content/groups/internal/</url>
```
maxdanilov, line 38: does it make sense to make this dynamic as well, with a default value set to `https://repository.nexus.camunda.cloud/content/groups/internal/`? (e.g. like we do with `server_id`). Not sure if it makes sense, just was thinking about it when seeing `server_id` dynamic.

maxdanilov (commented):
Looks good overall, leaving some comments for better understanding.

use-ci-nexus-cache/action.yml
maxdanilov: nvm, found answer in https://github.com/camunda/team-infrastructure/issues/282#issuecomment-1406390331

use-ci-nexus-cache/action.yml
chaima-mnsr: this is to make the action more generic, it could be that some teams are not using `camunda-nexus` as server id. maybe just `nexus` or `camunda_nexus`

maxdanilov: makes sense, thanks!

maxdanilov (approved):
🚀

PR #31, opened 2023-02-08
Add a new 'configure-pull-request' composite GitHub action

clementnero:
Related to https://github.com/camunda/team-infrastructure/issues/204

Add a new composite GitHub action to configure a Pull Request. It is part of the refactoring of the `maintenance_project.yml` workflow but the logic can be reused for other needs.

### DIFF
--- a/configure-pull-request/README.md
+++ b/configure-pull-request/README.md
@@ -9,7 +9,7 @@ This composite GHA can be used in any repository and must be triggered on `pull_
 ### Inputs
 | Input name           | Description                                               |
 |----------------------|-----------------------------------------------------------|
-| github-token         | A GitHub personal access token with sufficient scope (*) |
+| github-token         | A GitHub access token with sufficient scope (*) |
 | labels               | Comma-separated list of labels to set |
 | project-url          | URL of a GitHub project to which to add the PR |
 | reviewers            | Comma-separated list of users to set as reviewers |
@@ -32,7 +32,7 @@ jobs:
     steps:
     - uses: camunda/infra-global-github-actions/configure-pull-request@main
       with:
-        github-token: ${{ secrets.MY_PAT }}
+        github-token: ${{ secrets.A_GITHUB_TOKEN }}
         labels: label-1,label-2
         project-url: https://github.com/orgs/camunda/projects/project-id/
         reviewers: user1,user2
--- a/configure-pull-request/README.md
+++ b/configure-pull-request/README.md
@@ -1,0 +1,40 @@
+# configure-pull-request
+
+This composite Github Action (GHA) is aimed to be used by Camunda teams for configuring basic parameters of a GitHub Pull Request like labels, project and reviewers.
+
+## Usage
+
+This composite GHA can be used in any repository and must be triggered on `pull_request` events.
+
+### Inputs
+| Input name           | Description                                               |
+|----------------------|-----------------------------------------------------------|
+| github-token         | A GitHub personal access token with sufficient scope (*) |
+| labels               | Comma-separated list of labels to set |
+| project-url          | URL of a GitHub project to which to add the PR |
+| reviewers            | Comma-separated list of users to set as reviewers |
+| team-reviewers       | Comma-separated list of teams to set as reviewers |
+
+(*) minimum scope required:
+- repo
+- admin:org/read:org (required if `team-reviewers` input is set)
+- project (required if `project-url` input is set)
+
+### Workflow Example
+```yaml
+---
+name: example
+on:
+  pull_request:
+jobs:
+  configure-pr:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: camunda/infra-global-github-actions/configure-pull-request@main
+      with:
+        github-token: ${{ secrets.MY_PAT }}
+        labels: label-1,label-2
+        project-url: https://github.com/orgs/camunda/projects/project-id/
+        reviewsers: user1,user2
+        team-reviewers: team1,team2
+```
--- a/configure-pull-request/README.md
+++ b/configure-pull-request/README.md
@@ -35,6 +35,6 @@ jobs:
         github-token: ${{ secrets.MY_PAT }}
         labels: label-1,label-2
         project-url: https://github.com/orgs/camunda/projects/project-id/
-        reviewsers: user1,user2
+        reviewers: user1,user2
         team-reviewers: team1,team2
 ```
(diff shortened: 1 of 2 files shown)


configure-pull-request/README.md
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  configure-pr:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: camunda/infra-global-github-actions/configure-pull-request@main
+      with:
+        github-token: ${{ secrets.MY_PAT }}
+        labels: label-1,label-2
+        project-url: https://github.com/orgs/camunda/projects/project-id/
+        reviewsers: user1,user2
```
maxdanilov, line 38:
```suggestion
        reviewers: user1,user2
```

configure-pull-request/action.yml
```diff
@@ -0,0 +1,49 @@
+name: Configure Pull Request
+
+description: Sets the parameters of an existing Pull Request at once
+
+inputs:
+  github-token:
+    description: A GitHub personal access token with write access to the project
```
maxdanilov, line 7:
```suggestion
    description: A GitHub access token with write access to the project
```
A PAT is something bound to a user account; access tokens would include both PATs and dynamic tokens like the ones obtained with [tibdex/github-app-token](https://github.com/camunda/github-actions-recipes/blob/main/README.md#authenticate-from-a-github-app-via-tibdexgithub-app-token-usages-in-camunda)

configure-pull-request/action.yml
```diff
@@ -0,0 +1,49 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Set labels
+      uses: buildsville/add-remove-label@eeae411a9be2e173f2420e1644514edbecc4e835 # pin@v2.0.0
+      with:
+        labels: ${{ inputs.labels }}
+        token: ${{ inputs.github-token }}
+        type: add
+      if: ${{ inputs.labels != '' }}
```
maxdanilov, line 31: I would suggest moving the `if` condition right under `name` so it's immediately clear that this step is conditional

configure-pull-request/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Set project
+      uses: actions/add-to-project@960fbad431afda394cfcf8743445e741acd19e85 # pin@v0.4.0
+      with:
+        github-token: ${{ inputs.github-token }}
+        project-url: ${{ inputs.project-url }}
+      if: ${{ inputs.project-url != '' }}
```
maxdanilov, line 37: I would suggest moving the `if` condition right under `name` so it's immediately clear that this step is conditional

configure-pull-request/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Set reviewers (users)
+      uses: dannysauer/actions-assigner@ca2ca780928294ee5126986cc31db6c2ee7d69b0 # pin@v2.0.1
+      with:
+        reviewers: ${{ inputs.reviewers }}
+        token: ${{ inputs.github-token }}
+      if: ${{ inputs.reviewers != '' }}
```
maxdanilov, line 43: I would suggest moving the `if` condition right under `name` so it's immediately clear that this step is conditional

configure-pull-request/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Set reviewers (teams)
+      uses: dannysauer/actions-assigner@ca2ca780928294ee5126986cc31db6c2ee7d69b0 # pin@v2.0.1
+      with:
+        team-reviewers: ${{ inputs.team-reviewers }}
+        token: ${{ inputs.github-token }}
+      if: ${{ inputs.team-reviewers != '' }}
```
maxdanilov, line 49: I would suggest moving the `if` condition right under `name` so it's immediately clear that this step is conditional

configure-pull-request/action.yml
maxdanilov, line 7: I'd suggest adding explicit types for the inputs (e.g., `type: string`): https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-of-onworkflow_callinputs

Currently, GH ignores that in runtime, but this behavior may be useful in the future and provides a self-documenting code.

configure-pull-request/action.yml
```diff
@@ -0,0 +1,49 @@
...
+        project-url: ${{ inputs.project-url }}
+      if: ${{ inputs.project-url != '' }}
+    - name: Set reviewers (users)
+      uses: dannysauer/actions-assigner@ca2ca780928294ee5126986cc31db6c2ee7d69b0 # pin@v2.0.1
+      with:
+        reviewers: ${{ inputs.reviewers }}
+        token: ${{ inputs.github-token }}
+      if: ${{ inputs.reviewers != '' }}
+    - name: Set reviewers (teams)
+      uses: dannysauer/actions-assigner@ca2ca780928294ee5126986cc31db6c2ee7d69b0 # pin@v2.0.1
```
maxdanilov, line 45: will Renovate update this action in the future if a new version is released or will it be pinned to 2.0.1 permanently?

maxdanilov (changes requested):

configure-pull-request/action.yml
clementnero: Input type seems not to be allowed for composite action. I will try.

configure-pull-request/action.yml
clementnero: I am not familiar with Renovate configuration but it seems that digest can be updated. What is the Infra team strategy to deal with (small) external action ?

maxdanilov: We usually just specify the semver (i.e., `uses: dannysauer/actions-assigner@2.0.1`)

Renovate also supports hash pinning with following the updates: https://docs.renovatebot.com/modules/manager/github-actions/#additional-information

configure-pull-request/action.yml
maxdanilov: ah I didn't know. Then let's not do it.

clementnero: I have added the types for the inputs as suggested (it works, you were rights)

configure-pull-request/action.yml
clementnero: > We usually just specify the semver (i.e., `uses: dannysauer/actions-assigner@2.0.1`) and Renovate creates PR when a new version is released.
> 
> Renovate also supports hash pinning with following the updates: https://docs.renovatebot.com/modules/manager/github-actions/#additional-information

I will specify the semver for now. Thx for the link.

maxdanilov (approved):
Looks great, pretty lean!

PR #33, opened 2023-02-13
chore(gha) Adding additional inputs for docker build.

nwhinkle:
This PR will add some additional inputs for the `docker build and push` action and make some configurable with sensible defaults:

- build_context: Many images place dockerfiles and install scripts in subdirectories. This will default to root.
- build_allow: This allows configuring additional entitlements when building images that require this.
- buildx_driver: Allows you to specify which driver to use for the buildx builder.
- Builx_version: Allows you to specify a certain revision of the buildx action to use when building.


Langleu (approved):
looks good 👍 .
Had to test it out first since I wanted to make sure what happens if those inputs aren't supplied but it will just ignore them.
E.g. buildx will drop the `with` part since it's empty.

nwhinkle:
> looks good +1 . Had to test it out first since I wanted to make sure what happens if those inputs aren't supplied but it will just ignore them. E.g. buildx will drop the `with` part since it's empty.

Yes, I should have stated that in description. They'll be ignored if not supplied.

PR #34, opened 2023-02-13
Add a new `generate-github-app-token-from-vault-secrets` composite GHA

clementnero:
Related to https://github.com/camunda/team-infrastructure/issues/204

Add a new composite GitHub action to generate a GitHub token from Github App secrets (ID & Private Key) stored in Vault. It is part of the refactoring of the `maintenance_project.yml` workflow.

### DIFF
--- a/generate-github-app-token-from-vault-secrets/action.yml
+++ b/generate-github-app-token-from-vault-secrets/action.yml
@@ -1,0 +1,88 @@
+name: Generate a GitHub App token from secrets stored in Vault
+
+description: Generate a GitHub token from GitHub App secrets (ID & Private Key) stored in Vault
+
+inputs:
+  github-app-id-vault-key:
+    description: The key of the Vault secret storing the ID of the GitHub App
+    required: true
+    type: string
+  github-app-id-vault-path:
+    description: The path of the Vault secret storing the ID of the GitHub App
+    required: true
+    type: string
+  github-app-private-key-vault-key:
+    description: The key of the Vault secret storing the private key of the GitHub App
+    required: true
+    type: string
+  github-app-private-key-vault-path:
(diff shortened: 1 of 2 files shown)


generate-github-app-token-from-vault-secrets/action.yml
```diff
@@ -0,0 +1,88 @@
...
+  token:
+    description: The generated GitHub token
+    value: ${{ steps.github-token.outputs.token }}
+
+runs:
+  using: composite
+  steps:
+  - name: Import the GitHub App ID from vault
+    id: app-id
+    uses: hashicorp/vault-action@v2.4.3
```
maxdanilov, line 49:
```suggestion
    uses: hashicorp/vault-action@v2.5.0
```

let's use the recent version of it

generate-github-app-token-from-vault-secrets/action.yml
```diff
@@ -0,0 +1,88 @@
...
+          format(
+            '{0} {1} | {2};',
+            inputs.github-app-id-vault-path,
+            inputs.github-app-id-vault-key,
+            'id'
+          ) 
+        }}
+  - name: Import the GitHub App secret key from vault
+    id: app-private-key
+    uses: hashicorp/vault-action@v2.4.3
```
maxdanilov, line 67:
```suggestion
    uses: hashicorp/vault-action@v2.5.0
```

let's use the recent version of it

maxdanilov (approved):
Looks good to me, a really nice and clean implementation.
Just one suggestion to use the current version of the `vault-action`

generate-github-app-token-from-vault-secrets/action.yml
clementnero: Good point!

PR #50, opened 2023-05-23
Don't use Python in submit-build-action

cmur2:
We discovered that GHA self-hosted runner images based on Ubuntu 22.04 do not contain `python` by default anymore - find the background here: https://github.com/actions/actions-runner-controller/pull/2050

We only need Python here for input validation (nicer to code than in Bash) but it can be done in Bash as well which avoids having to install Python just for this.

Example runs:

- failed run when one input is too long (I reduced the limit to 2 chars): https://github.com/camunda/infra-core/actions/runs/5058585353/jobs/9078888781?pr=5855
- successful run with proper inputs: https://github.com/camunda/infra-core/actions/runs/5058585353/jobs/9078992674?pr=5855

### DIFF
--- a/submit-build-status/action.yml
+++ b/submit-build-status/action.yml
@@ -33,17 +33,26 @@ runs:
       echo "User description: ${{ inputs.user_description }}"
 
   - name: Validate inputs
-    shell: python
+    shell: bash
+    env:
+      BUILD_STATUS: "${{ inputs.build_status }}"
+      USER_REASON: "${{ inputs.user_reason }}"
+      USER_DESCRIPTION: "${{ inputs.user_description }}"
     run: |
-      if '${{ inputs.build_status }}' == '':
-        print('Need to specify a build status that is not empty!')
-        sys.exit(1)
-      if len('${{ inputs.user_reason }}') > 200:
-        print('Specified user reason has more than 200 characters!')
-        sys.exit(1)
-      if len('${{ inputs.user_description }}') > 1000:
-        print('Specified user description has more than 1000 characters!')
-        sys.exit(1)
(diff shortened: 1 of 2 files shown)


submit-build-status/action.yml
```diff
@@ -33,17 +33,25 @@ runs:
...
-      if '${{ inputs.build_status }}' == '':
-        print('Need to specify a build status that is not empty!')
-        sys.exit(1)
-      if len('${{ inputs.user_reason }}') > 200:
-        print('Specified user reason has more than 200 characters!')
-        sys.exit(1)
-      if len('${{ inputs.user_description }}') > 1000:
-        print('Specified user description has more than 1000 characters!')
-        sys.exit(1)
+      if [ "${{ inputs.build_status }}" == "" ]; then
```
Langleu, line 41: why do we define the other inputs as env var but this one we template the gha way?
Maybe for consistency we could move it to be a env var as well?

submit-build-status/action.yml
```diff
@@ -57,7 +65,6 @@ runs:
   - name: Submit build status to CI Analytics
     shell: bash
     env:
-      # TODO: change to prod
```
Langleu, line 60: ^^

Langleu (approved):
looks good 👍 
just one small remark to consistency but also fine to keep it that way.

submit-build-status/action.yml
cmur2: Yea can do that too for consistency.

Background: the check for an empty string also works when comparing to literals but determining the string length requires a variable beforehand.

submit-build-status/action.yml
cmur2: Lalala 🙈

PR #62, opened 2023-07-11
chore(gha): add common-tooling action for easier use with self-hosted runners

Langleu:
related to https://github.com/camunda/team-infrastructure/issues/389

The action pretty much just copies the inputs from the existing actions and prefixes them with whatever the source is. Meaning you're in full control to overwrite anything.
Before running a check is done to see what's available to potentially skip installing a software. Requires `overwrite = false`. Default is to overwrite to have the same versions on `hosted` and `self-hosted` since on `hosted` everything is available.

Creates `common-tooling` action with some defaults on.
Node - 16 lts --> most used by GitHub Actions
Python - 3.11
Java - 17 LTS
Rest is latest where possible.

Added a workflow to test the action with different parameters.
In the case software is already available, the run is a lot quicker, see GHA vs self-hosted or any subsequent run.

### DIFF
--- a/.github/workflows/test-common-tooling.yml
+++ b/.github/workflows/test-common-tooling.yml
@@ -1,0 +1,75 @@
+---
+name: Test Common Tooling
+
+on:
+  push:
+
+jobs:
+  test-common-tooling:
+    strategy:
+      matrix:
+        label: ["ubuntu-latest", "gcp-core-2-default"]
+    runs-on: ${{ matrix.label }}
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./common-tooling
+      with:
+        secrets: ${{ toJSON(secrets) }}
+    - run: |
+        if ! command -v node &> /dev/null
+        then
+            echo "node could not be found"
+            exit 1
+        else
+            node --version
+        fi
+        if ! command -v yarn &> /dev/null
+        then
+            echo "yarn could not be found"
+            exit 1
+        else
+            yarn --version
+        fi
+        if ! command -v docker buildx &> /dev/null
+        then
+            echo "docker buildx could not be found"
+            exit 1
+        else
+            docker buildx version
+        fi
+        if ! command -v java &> /dev/null
+        then
+            echo "java could not be found"
+            exit 1
+        else
+            java --version
+        fi
+        if ! command -v mvn &> /dev/null
+        then
+            echo "mvn could not be found"
+            exit 1
+        else
+            mvn --version
+        fi
+        if ! command -v python &> /dev/null
+        then
+            echo "python could not be found"
+            exit 1
+        else
+            python --version
+        fi
+    - uses: ./common-tooling
+      with:
+        secrets: ${{ toJSON(secrets) }}
+        node-enabled: "false"
+        yarn-enabled: "false"
+        python-enabled: "false"
+        java-enabled: "false"
+        qemu-enabled: "false"
+        buildx-enabled: "false"
+    - uses: ./common-tooling
+      with:
+        secrets: ${{ toJSON(secrets) }}
+        overwrite: "false"
+    - uses: ./common-tooling
--- a/.github/workflows/test-common-tooling.yml
+++ b/.github/workflows/test-common-tooling.yml
@@ -9,16 +9,17 @@ on:
 jobs:
   test-common-tooling:
     strategy:
+      fail-fast: false
       matrix:
         label: ["ubuntu-latest", "gcp-core-2-default"]
     runs-on: ${{ matrix.label }}
     timeout-minutes: 10
     steps:
     - uses: actions/checkout@v3
-    - uses: ./common-tooling
-      with:
-        secrets: ${{ toJSON(secrets) }}
-    - run: |
+    - name: Install if required common software tooling
+      uses: ./common-tooling
+    - name: Check that installed software is available
+      run: |
         if ! command -v node &> /dev/null
         then
             echo "node could not be found"
@@ -33,7 +34,8 @@ jobs:
         else
             yarn --version
         fi
-        if ! command -v docker buildx &> /dev/null
+        docker_buildx_count=$(docker buildx 2>&1 | grep -c "docker buildx")
+        if [[ $docker_buildx_count -eq 0 ]]
         then
             echo "docker buildx could not be found"
             exit 1
@@ -61,17 +63,24 @@ jobs:
         else
             python --version
         fi
-    - uses: ./common-tooling
+    - name: Install nothing, just for testing
+      uses: ./common-tooling
       with:
-        secrets: ${{ toJSON(secrets) }}
         node-enabled: "false"
         yarn-enabled: "false"
         python-enabled: "false"
         java-enabled: "false"
         qemu-enabled: "false"
         buildx-enabled: "false"
-    - uses: ./common-tooling
+    - name: Install and overwrite common software tooling
+      uses: ./common-tooling
       with:
-        secrets: ${{ toJSON(secrets) }}
-        overwrite: "false"
-    - uses: ./common-tooling
+        overwrite: "true"
+    - name: Install and overwrite common software tooling
+      uses: ./common-tooling
+      with:
+        node-version: 20 # see https://github.com/actions/setup-node#supported-version-syntax for supported syntax
+        java-version: 20 # see https://github.com/marketplace/actions/setup-java-jdk#supported-version-syntax for supported syntax
+        java-distribution: adopt # see https://github.com/marketplace/actions/setup-java-jdk#supported-distributions for supported distros
+        python-version: 3 # see https://github.com/actions/setup-python#supported-version-syntax for supported syntax
+        overwrite: 'true'
--- a/.github/workflows/test-common-tooling.yml
+++ b/.github/workflows/test-common-tooling.yml
@@ -2,7 +2,9 @@
 name: Test Common Tooling
 
 on:
-  push:
+  pull_request:
+    paths:
+    - 'common-tooling/action.yml'
 
 jobs:
   test-common-tooling:
--- a/common-tooling/README.md
+++ b/common-tooling/README.md
@@ -2,11 +2,34 @@
 
 This composite Github Action (GHA) is aimed to be used by Camunda teams for configuring either hosted or self-hosted runners with some basic software. Especially useful to have the basics available on self-hosted runners since they come with close to no software.
 
+## Default Software
+- Buildx - latest
+- JDK - 17
+- Maven - 3.9
+- Node - 16
+- Python - 3.11
+- Qemu - for multi arch builds (has to be enabled via `qemu-enabled: true`)
+- Yarn - 1.22
+
+## Used Actions
+See following actions for further information on available inputs and their usages.
+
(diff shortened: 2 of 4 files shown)


.github/workflows/test-common-tooling.yml
```diff
@@ -0,0 +1,77 @@
...
+jobs:
+  test-common-tooling:
+    strategy:
+      matrix:
+        label: ["ubuntu-latest", "gcp-core-2-default"]
+    runs-on: ${{ matrix.label }}
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./common-tooling
```
maxdanilov, line 18: I suggest giving the steps the human-readable names for identification as well, e.g.:
```suggestion
    - uses: ./common-tooling
      name: Install defaults and configure maven-settings.xml
```

.github/workflows/test-common-tooling.yml
```diff
@@ -0,0 +1,77 @@
...
+            mvn --version
+        fi
+        if ! command -v python &> /dev/null
+        then
+            echo "python could not be found"
+            exit 1
+        else
+            python --version
+        fi
+    - uses: ./common-tooling
```
maxdanilov, line 64: Same for this step and the others below (`name`)

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+    description: 'QEMU static binaries Docker image (e.g. tonistiigi/binfmt:latest)'
+    default: 'tonistiigi/binfmt:latest'
+    required: false
+  qemu-platforms:
+    description: 'Platforms to install (e.g. arm64,riscv64,arm)'
+    default: 'all'
+    required: false
+  qemu-enabled:
+    description: 'Whether to install qemu or not'
+    default: 'true'
```
maxdanilov, line 94: Do many teams need it by default? Shouldn't it be `false`?

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+
+      if ! command -v yarn &> /dev/null
+      then
+          echo "yarn_missing=true" >> "$GITHUB_ENV"
+      fi
+
+      if ! command -v java &> /dev/null
+      then
+          echo "java_missing=true" >> "$GITHUB_ENV"
+      fi
```
maxdanilov, line 192: maybe add something like `echo $GITHUB_ENV` here to see what tools are missing - hard to see that info in the GHA logs.

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+  - name: Install Yarn
+    if: ${{ inputs.yarn-enabled == 'true' && ( env.yarn_missing == 'true' || inputs.overwrite == 'true' ) }}
+    shell: bash
+    run: |
+      curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+      echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+      sudo apt update && sudo apt install yarn -y
+      export PATH="$PATH:`yarn global bin`"
+  # Install node
+  - uses: actions/setup-node@v3
```
maxdanilov, line 228: `name` is missing here.

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+        [{
+          "id": "camunda-nexus",
+          "username": "${{ steps.secrets.outputs.NEXUS_USR }}",
+          "password": "${{ steps.secrets.outputs.NEXUS_PSW }}"
+        }]
+      mirrors: '[{"url": "https://repository.nexus.camunda.cloud/content/groups/internal/", "id": "camunda-nexus", "mirrorOf": "*", "name": "camunda Nexus"}]'
+  # Change Timezone to Berlin
+  - shell: bash
+    if: ${{ inputs.berlin-timezone == 'true' }}
+    run: |
```
maxdanilov, line 279: `name` is missing for this step.

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+      sudo rm -rf /etc/localtime
+      sudo ln -s /usr/share/zoneinfo/Europe/Berlin /etc/localtime
+  # Install Python, removed on self-hosted runners
+  - uses: actions/setup-python@v4
+    if: ${{ inputs.python-enabled == 'true' && ( env.python_missing == 'true' || inputs.overwrite == 'true' ) }}
```
maxdanilov, line 284: `name` is missing here.

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+  python-allow-prereleases:
+    description: "When 'true', a version range passed to 'python-version' input will match prerelease versions if no GA versions are found. Only 'x.y' version range is supported for CPython."
+    default: "false"
+  python-enabled:
+    description: 'Whether to install python or not'
+    default: 'true'
+  # misc
+  overwrite:
+    description: 'Defines whether on hosted runners the present version should be overwritten'
+    default: 'true'
```
maxdanilov, line 152: should it rather be `true` so the default behaviour in SaaS runners takes less runtime?

maxdanilov (changes requested):

.github/workflows/test-common-tooling.yml
maxdanilov: Though I guess it's not very usable in the logs now (https://github.com/actions/runner/issues/1877) still adding a bit readability be good.

.github/workflows/test-common-tooling.yml
```diff
@@ -0,0 +1,77 @@
...
+        yarn-enabled: "false"
+        python-enabled: "false"
+        java-enabled: "false"
+        qemu-enabled: "false"
+        buildx-enabled: "false"
+    - uses: ./common-tooling
+      with:
+        secrets: ${{ toJSON(secrets) }}
+        overwrite: "false"
+    - uses: ./common-tooling
```
maxdanilov, line 77: maybe the first combination should be `./common-tooling` with `overwrite: false`, so in the SaaS runner we can verify it's quick (under a few seconds) as a no-op with the action defaults. Right now it's hard to verify it.

maxdanilov (changes requested):

common-tooling/README.md
```diff
@@ -0,0 +1,81 @@
...
+    - uses: camunda/infra-global-github-actions/common-tooling@main
+    # install any defaults and configure maven-settings.xml
+    - uses: camunda/infra-global-github-actions/common-tooling@main
+      with:
+        secrets: ${{ toJSON(secrets) }}
+    # only install missing software components
+    - uses: camunda/infra-global-github-actions/common-tooling@main
+      with:
+        overwrite: "false"
+```
```
maxdanilov, line 81: I'd suggest adding more examples here of config options for specific tooling passed (like version of Node etc.), something simple yet giving the users the taste of it.

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+      githubServer: false
+      servers: |
+        [{
+          "id": "camunda-nexus",
+          "username": "${{ steps.secrets.outputs.NEXUS_USR }}",
+          "password": "${{ steps.secrets.outputs.NEXUS_PSW }}"
+        }]
+      mirrors: '[{"url": "https://repository.nexus.camunda.cloud/content/groups/internal/", "id": "camunda-nexus", "mirrorOf": "*", "name": "camunda Nexus"}]'
+  # Change Timezone to Berlin
+  - shell: bash
```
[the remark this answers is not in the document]
Langleu: don't have to, just a common thing in AP an Optimize at least.

common-tooling/action.yml
Langleu: the idea was of a drop-in replacement to be equal for either hosted or self-hosted. Meaning if you run it on either, you should expect to get the same results but I'm fine changing that up.

common-tooling/action.yml
```diff
@@ -0,0 +1,302 @@
...
+      token: ${{ inputs.node-token }}
+      cache: ${{ inputs.node-cache }}
+      cache-dependency-path: ${{ inputs.node-cache-dependency-path }}
+  # Setup Maven
+  - name: Import secrets
+    if: ${{ inputs.secrets != '' && inputs.java-enabled == 'true' && ( env.java_missing == 'true' || inputs.overwrite == 'true' ) }}
+    id: secrets
+    uses: hashicorp/vault-action@v2.7.0
+    with:
+      url: ${{ fromJSON(inputs.secrets).VAULT_ADDR }}
```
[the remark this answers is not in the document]
Langleu: doesn't make much difference whether you do it outside or inside.
Yes, question comes down to whether just installing tools or also do some configuration.

common-tooling/action.yml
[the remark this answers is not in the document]
maxdanilov: then maybe it's better to drop it?

common-tooling/action.yml
maxdanilov: I'd see that as an action that makes SH behave the same as SaaS (i.e. on SaaS it's a no-op that installs nothing and on SH it installs the tooling missing there to match default `ubuntu-latest` SaaS one)

common-tooling/action.yml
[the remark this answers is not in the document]
Langleu: dropped

common-tooling/README.md
```diff
@@ -0,0 +1,81 @@
...
+```yaml
+---
+name: example
+on:
+  pull_request:
+jobs:
+  some-job:
+    runs-on: ubuntu-latest
+    steps:
+    # install any defaults
```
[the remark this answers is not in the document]
Langleu: added

common-tooling/README.md
```diff
@@ -0,0 +1,81 @@
...
+    - uses: camunda/infra-global-github-actions/common-tooling@main
+    # install any defaults and configure maven-settings.xml
+    - uses: camunda/infra-global-github-actions/common-tooling@main
+      with:
+        secrets: ${{ toJSON(secrets) }}
```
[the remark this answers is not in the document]
Langleu: dropped due to removal of the vault-action

common-tooling/action.yml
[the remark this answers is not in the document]
Langleu: dropped the configuration --> removal of vault as well

common-tooling/action.yml
Langleu: put it to `false` by default.

common-tooling/action.yml
Langleu: added

common-tooling/action.yml
Langleu: dropped the timezone

common-tooling/action.yml
Langleu: switched to `false` as default.

.github/workflows/test-common-tooling.yml
Langleu: switched it up

common-tooling/README.md
Langleu: added another example with versions used and comments to their syntax.
Additionally added in the readme the source actions for quick access and usage to their inputs.

maxdanilov (approved):
Looks great, thank you! 🚀

PR #73, opened 2023-08-28
chore(preview-env): Fix deployment status

clementnero:
This PR fixes the status of GitHub deployments when `preview-env/create` GHA fails, and enables:
- a more readable status for the end-user. If the deployment workflow fails, the environment is marked as failed in the PR and no link is shown.
- the cleanup workflow to avoid missing environments falsely considered inactive when a deployment job is cancelled
- to extend the logic later on to apply more refined shutdown rules depending on the actual status of the deployment (e.g. quicker shutdown of an environment in case of error)

In a composite action, `job.status` is never updated and always takes the value `success`. So even if the GHA fails, the GitHub deployment status was always set to `success` which leads to inconsistencies. **`github.action_status`** should be used instead.

The GitHub action [bobheadxi/deployments ](https://github.com/bobheadxi/deployments) considers that a `cancelled` status [corresponds](https://github.com/bobheadxi/deployments/blob/8a87bc300d207ce86142a8080e9b8fabec463494/src/steps/finish.ts#L47) to an inactive deployment. In our case, a deployment workflow can be canceled due to a timeout or a user action. Resources are not deleted, so the GitHub deployment must be considered a failure and therefore still live even if in error. The fix explicitly set the status to failure if the GHA is cancelled.

Tested on Cawemo in a dry-run mode: 
image


preview-env/clean/action.yml
```diff
@@ -58,6 +58,9 @@ inputs:
 
       ⏯️ **Want to keep preview environment(s) for longer ?**
       Simply push a new commit or restart the deployment [workflow]({checks-url}) to **postpone the shutdown to `{ttl-days} day(s)` from now**.
+
+      🔁 **Want to keep preview environment(s) permanently ?**
+      Consider deploying a static environment associated with a long-running branch.
```
Langleu, line 63: what exactly do you mean by it?
It sounds like we have "stable" preview environments but we don't have that.

Langleu (changes requested):
see comment above

preview-env/clean/action.yml
clementnero: Indeed it is not very clear. I was actually thinking of stable branches when I wrote long-running branches. The idea was to anticipate the needs of teams to deploy environments for their stable branches and point them towards the right solution and avoid them hijacking the use of dynamic preview environments as [Operate](https://github.com/camunda/operate/pull/4741) has done. Operate is a special case because preview environments are already activated for stable branches (this PR needs to be closed without triggering a teardown).

I could change to:
```suggestion
      🔁 **Want to keep preview environment permanently for a stable branch?**
      Consider deploying a static environment and reach Infra team for support.
```

Or I can just add as a reminder in the announcement. 

What do you think ?

Langleu: Your suggestion is more clear 👍.
Yeah always wondered what that Operate PR was about since `stable/**` is already supported at them.

Langleu: but I'm still not sure whether to keep it.
E.g. Web-Modeler asked recently about having a branch deployed for a longer period, which they I guess would consider stable in that case but I don't think we should create temporary exceptions for those "stable" branches.
`stable/**` I see as an exclusion since we already have those as exceptions in Operate and Tasklist and the images are tagged accordingly.

the `stable/**` should already work for them out of the box, so at them it's simply a misuse of the preview-environments.

Langleu: so overall, would remove it to not create a potential of us doing exceptions all the time for feature branches that need to "live" longer.

Or change it to something like `Want to keep a preview environment for a release branch?`

clementnero: I see 👍 . Let's delete it for now. We'll adjust it later if there is a need.

Langleu (approved):

PR #77, opened 2023-09-06
chore(deployments): use semver instead of hash versioning

maxdanilov:
List of releases: https://github.com/bobheadxi/deployments/releases


cmur2 (approved):
I agree that in this case the readability is more important than pinning to the immutable hash (which is usually done for security if you audit the updates).

PR #92, opened 2023-09-18
chore(preview-env/create): Add a retry logic to the `argocd` `wait` command

clementnero:
Related to ask-infra request https://camunda.slack.com/archives/C5AHF1D8T/p1694771504886029.

Permission denied issues are encountered by Operate when deploying a preview environment even though the token used has the right permissions.

It's not related to the Dex crash loop as mentioned in this [comment](https://camunda.slack.com/archives/CHY2S7KDJ/p1694430900442579?thread_ts=1694428700.790319&cid=CHY2S7KDJ). Dex is not part of the authorization validation process.

The most likely hypothesis is a consistency problem between the creation of the app and the execution of the wait command immediately afterwards, as suggested [here](https://camunda.slack.com/archives/CTYKBESCE/p1695026795048189?thread_ts=1695026504.084219&cid=CTYKBESCE), but I couldn't reproduce this scenario in dev.

As a first iteration, we can set up a retry mechanism and see if this solves the problem.

PR has been [validated](https://github.com/camunda/operate/actions/runs/6227962067) on this [test](https://github.com/camunda/operate/tree/infra-request-test-do-not-merge) branch.

Once merged:
- [ ] Delete the [test](https://github.com/camunda/operate/tree/infra-request-test-do-not-merge) branch and associated argocd app

### DIFF
--- a/preview-env/create/action.yml
+++ b/preview-env/create/action.yml
@@ -98,7 +98,7 @@ runs:
         fi
         echo "Retry $i ..."
         ((i++))
-        sleep 1
+        sleep 5
       done
       echo "::endgroup::"
     env:
--- a/preview-env/create/action.yml
+++ b/preview-env/create/action.yml
@@ -80,7 +80,14 @@ runs:
     shell: bash
     run: |-
       echo "::group::wait for sync"
-      argocd app wait ${{ inputs.app_name }} --timeout 1200 --health
+      i=1
+      until argocd app wait ${{ inputs.app_name }} --timeout 1200 --health \
+        || [ $i -gt 5 ];
+      do
+        echo Retry $i ...
+        sleep 1
+        ((i++))
+      done
       echo "::endgroup::"
     env:
       ARGOCD_AUTH_TOKEN: ${{ inputs.argocd_token }}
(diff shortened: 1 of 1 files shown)


preview-env/create/action.yml
```diff
@@ -80,7 +80,14 @@ runs:
     shell: bash
     run: |-
       echo "::group::wait for sync"
-      argocd app wait ${{ inputs.app_name }} --timeout 1200 --health
+      i=1
+      until argocd app wait ${{ inputs.app_name }} --timeout 1200 --health \
```
cmur2, line 84: The `until` will retry the `argocd app wait ${{ inputs.app_name }} --timeout 1200 --health` whenever that exits with non-zero error code, if I understand correctly.

This can happen for multiple reasons (IIRC, we should double check):

- authentication problem (intended, this is what the PR wants to address)
- app doesn't become healthy within the timeout e.g. due to a persistent developer coding error (unintended, could lead also to cost increase since GHA workflow will run 5-times longer potentially)

Maybe there is a way to distinguish the two?

clementnero: Good point 👍 
ArgoCD seems to exit with a generic code (`20`), so `stderr` needs to be parsed for `code = PermissionDenied`.

clementnero:
I have increased retry interval to 5s.

cmur2 (commented):
Nice, thanks for addressing the edge cases and doing testing :+1:

PR #111, opened 2023-11-03
chore(download-center-upload): Add action input to determine if repo …

andromaqui:
…name should be in the upload path

Tested by running the optimize release workflow and linking the action in this branch:
Pr: https://github.com/camunda/camunda-optimize/pull/9278

TEST 1:
repo name override by 'optimize
commit: https://github.com/camunda/camunda-optimize/pull/9278/commits/5b46cfdb223f67696a2f26545f0fd9ec3ddfe1c6
Result: https://dev.downloads.camunda.cloud/enterprise-release/optimize/9.99.0/
Workflow run: https://github.com/camunda/camunda-optimize/actions/runs/6745186903

TEST 2:
repo name not set
commit: https://github.com/camunda/camunda-optimize/pull/9278/commits/05db0d19f7e547ba37da786fc2301d193d8191d1
Result: https://dev.downloads.camunda.cloud/enterprise-release/camunda-optimize/8.888.0/
Workflow run: https://github.com/camunda/camunda-optimize/actions/runs/6745436645


download-center-upload/action.yml
```diff
@@ -25,6 +25,10 @@ inputs:
   artifact_file:
     description: Artifact(s) to upload. E.g. can contain a wildcard (*) or space separated string
     required: true
+  use_repo_name:
```
maxdanilov, line 28: I would suggest instead to have a parameter `repo_name` which by default falls back to the repo name (e.g. `camunda-optimize`) but can also be overwritten (with `optimize` for example). The if-else switch overwise is too verbose.

download-center-upload/action.yml
```diff
@@ -25,6 +25,10 @@ inputs:
...
+    description: Variable which indicates whether the name of the repository should be included in the upload subpath.
+    default: 'true'
+    required: true
```
maxdanilov, line 31: Did you check that this works as expected? e.g. when `use_dc_repo` is not specified, does it work or crash?

maxdanilov (changes requested):

download-center-upload/action.yml
andromaqui: I think if the repo name defaults to camunda-optimize we would still need an if else statement to check whether we are currently uploading Optimize files right? Because otherwise all other teams which use this action would upload in camunda-optimize so there would have to be some sort of logic in place to see whether we need to replace this default value

andromaqui: yeah that makes sense actually 👍🏻

download-center-upload/README.md
```diff
@@ -13,16 +13,16 @@ This action is owned by Infra team.
...
+| **sub_version**      | Patch version of the artifact.                                                                                            | :x:                |         |                        |                                   |
+| **artifact_subpath** | Sub-path of the artifact in Download center.                                                                              | :x:                |         |                        | should not start or end with `/`. |
+| **artifact_file**    | The artifact(s). Can be a wildcard (*) or space separated list or single string                                           | :heavy_check_mark: |         |                        |                                   |
+| **use_repo_name**    | Boolean value, if 'true' it included the repository name on the upload path                                               | :x:                | `true`  |  `true`  `false`       |                                   |
```
maxdanilov, line 25: the readme needs to be updated, it's not `use_repo_name` anymore

maxdanilov (changes requested):
The code looks good, yet the readme needs to be adjusted.

I'd also recommend to extend the readme to provide more examples of how the new `repo_name` parameter influences the target path of the artifacts

andromaqui:
Thank you! I added an additional example and modified the docs to include how to `repo_name ` can change the upload path.

maxdanilov (approved):
Looks good, thank you!

PR #113, opened 2023-11-14
chore(argocd-deploy): Preview Environment's ArgoCD app creation proce…

andromaqui:
…dure is more reliable


preview-env/create/action.yml
```diff
@@ -69,9 +69,40 @@ runs:
...
+        # Record the start time
+        start_time=$(date +%s)
+
+        # Wait for the app to be registered
+        while true; do
+          echo "Printing app info: "
+          app_info=$(argocd app get ${{ inputs.app_name }} --output json 2>/dev/null)
+          echo "$app_info"
+
+          if [ $? -eq 0 ]; then
```
Langleu, line 84: $? always looks at the last executed command, in that case `echo "$app_info"`.
echo is typically always executing successfully, doesn't matter whether `app_info=$(...)` will fail or not, it won't influence the echo command.

At least I tried it out and app_info was having a return code 20 but the following echo would just print nothing and be a 0

preview-env/create/action.yml
```diff
@@ -69,9 +69,40 @@ runs:
   - name: Deploy
     shell: bash
     run: >-
-      argocd app create ${{ inputs.argocd_arguments }}
+      if argocd app create ${{ inputs.argocd_arguments }}; then
```
Langleu, line 72: not really a fan of it being as part of a if statement, maybe run it and follow up with `$?`

Langleu (changes requested):
Left some comments and I think you're also missing the part about trying to create the app again.

from the desired outcome
```
after doing argocd app create, the script checks if the app has indeed been created and if not, tries again.
```

Currently it tries to create and either it works or fails there already.

Langleu:
my suggestion to the app create would be to try it 3 times and if it still can't create it, fail.

maxdanilov:
From my point of view, it should be: 
- 1. try to create 
- 2. wait for a few seconds (important from the eventual consistency point of view) 
- 3. check if created:
   - wait a few seconds and try 1 again if not succeeded (up to 3 times)

andromaqui:
Oh I am sorry I think I have therefore misunderstood the ticket potentially. 

> the script checks if the app has indeed been created and if not, tries again.

With "tries again" we therefore mean that the script tries again to create the app?

preview-env/create/action.yml
andromaqui: true echo is also a command 🤦‍♀️

andromaqui: I think it makes sense to check that app_info is not empty and that the exit code is 0. That way we know that fetching the app information didnt fail and that there is information associated to the app. Not sure if it is a bit redundant tho 🤔

preview-env/create/action.yml
```diff
@@ -68,10 +68,38 @@ runs:
...
+          argocd app create ${{ inputs.argocd_arguments }}
+          if [ $? -eq 0 ]; then
+            echo "Creating Argo CD app with name ${{ inputs.app_name }}"
+            sleep 60
 
-      argocd app sync ${{ inputs.app_name }} --async --force || true
+            # Fetch app information to make sure the app is registered
+            app_info=$(argocd app get ${{ inputs.app_name }} --output json 2>/dev/null)
+            get_status_status=$?
+            if [ $get_status_status -eq 0 ] && [ -n "$app_info" ]; then
```
Langleu, line 82: since you're immediately using `get_status_status`, you can just skip the variable assignment.

preview-env/create/action.yml
```diff
@@ -68,10 +68,38 @@ runs:
...
+              echo "App Info: $app_info"
```
Langleu, line 83: I'd suggest wrapping this `echo` statement in a group so it's collapsed in the workflow log and can be uncollapsed if needed. Makes the output more readable.

```
echo "::group::app info"
echo "App Info: $app_info"
echo "::endgroup::"
```

Langleu (approved):
Overall looks good 👍.
The for loop should work as expected as `--upsert` takes care of exiting as `0` on exiting apps.
So we try 3 times to create the app  / to make sure it's registered.
I agree with the sync adjustments of you as it should not happen anymore.

Left some small adjustments but more on a visual side.

preview-env/create/action.yml
Langleu: nothing about `--upsert`, was just an observation on how / why the for loop works.
`--upsert` is part of `${{ inputs.argocd_arguments }}` meaning --> `argocd app create ${{ inputs.argocd_arguments }}`.

`--upsert` overrides the existing app, so in case the app exists --> overwritten, if it doesn't exist --> created.
Either way the return code is `0`, which is important for your for loop to work in case the app registration didn't work immediately.

Nothing to change / adjust.

andromaqui: ah that makes sense 👍🏻

andromaqui: logs look good now https://github.com/camunda/operate/actions/runs/6877333070/job/18848699317

Langleu: sweet :rocket:, looks a lot more clean now. Thank you!

preview-env/create/action.yml
```diff
@@ -68,10 +68,39 @@ runs:
...
     shell: bash
-    run: >-
-      argocd app create ${{ inputs.argocd_arguments }}
+    run: |
+      max_retries=3
+      for attempt in $(seq "$max_retries"); do
+          argocd app create ${{ inputs.argocd_arguments }}
+          if [ $? -eq 0 ]; then
+            echo "Creating Argo CD app with name ${{ inputs.app_name }}"
+            sleep 60
```
maxdanilov, line 77: @andromaqui sorry for jumping on it, but why we sleep for a whole minute here? Eventual consistency app state convergence on the Argo side would be much faster than this, I don't see the need to sleep for a whole minute always, especially we have a retry mechanism in place. This makes a pipeline a whole minute longer for in my opinion not a good enough reason, negatively impacting user experience and increasing GHA costs.

I'd recommend something more moderate instead:
```suggestion
            sleep 10
```

andromaqui: yeah you are right since we retry 3 times maybe it is not very necessay, I will address this 👍🏻

PR #143, opened 2024-01-26
chore(gha): Implement a reusable workflow which allows retriggering g…

andromaqui:
…ha runs


rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
+name: Retrigger failed run action
+
+on:
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: The id of the run which we want to retrigger
+        required: true
+      error_message:
+        description: Custom error message to look for in the logs.
```
maxdanilov, line 10: For consistency with the other parameters' description fields:
```suggestion
        description: Custom error message to look for in the logs
```

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+      repository:
+        description: Github repository
+        required: true
+jobs:
+  rerun:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Import Secrets
+        id: secrets
+        uses: hashicorp/vault-action@v2.7.4
```
maxdanilov, line 21: the latest version is [2.8.0](https://github.com/hashicorp/vault-action/releases):
```suggestion
        uses: hashicorp/vault-action@v2.8.0
```

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+          # Fetch the logs from the previous run
+          logs=$(gh run view ${{ inputs.run_id }} -R ${{ inputs.repository }} --log)
+          echo "Logs coming from run with id: ${{ inputs.run_id }}"
+          echo "$logs"
+          echo "end logs"
+          # Check if the desired string is present in the logs
+          if echo "$logs" | grep -F "${{ inputs.error_message }}" >/dev/null; then
+            count=$(echo "$logs" | grep -F "${{ inputs.error_message }}" | wc -l)
+            echo "count: $count"
+            if [ "$count" -gt 2 ]; then
```
maxdanilov, line 54: Wny only a single occurence of the message is not sufficient to trigger a return? `"$count" -gt 2` means it will only trigger when there's 3 or more line matches, why is that?

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+          echo ${{ steps.github-token.outputs.token }} | gh auth login --with-token
+          echo "Fetching logs for run with id: ${{ inputs.run_id }} from ${{ inputs.repository }} repository."
+          # Fetch the logs from the previous run
+          logs=$(gh run view ${{ inputs.run_id }} -R ${{ inputs.repository }} --log)
+          echo "Logs coming from run with id: ${{ inputs.run_id }}"
+          echo "$logs"
+          echo "end logs"
+          # Check if the desired string is present in the logs
+          if echo "$logs" | grep -F "${{ inputs.error_message }}" >/dev/null; then
+            count=$(echo "$logs" | grep -F "${{ inputs.error_message }}" | wc -l)
```
maxdanilov, line 52: from the runtime efficiency, why there's a need to do `echo $logs | grep` twice (line 51 and 52)?
The way I see it, you can do this once before the `if`:
```shell
count=$(echo "$logs" | grep -F "${{ inputs.error_message }}" | wc -l)
```

and then only execute the code inside `if` when `count` is above 0 (meaning the string is present):

```shell
            if [ "$count" -gt 0 ]; then
```

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+            echo "count: $count"
+            if [ "$count" -gt 2 ]; then
+                echo "::set-output name=rerun::true"
```
maxdanilov, line 55: `set-output` are deprecated directives and should not be used anymore: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+              echo "Error message $error_message not found in logs"
+          fi
+
+      - name: Rerun run with id ${{ inputs.run_id }} for ${{ inputs.repository }} repository
+        if: steps.check_logs.outputs.rerun == 'true'
+        env:
+          GH_DEBUG: api
+        run: |
+          echo ${{ steps.github-token.outputs.token }} | gh auth login --with-token
+          gh run watch ${{ inputs.run_id }} -R ${{ inputs.repository }} > /dev/null 2>&1
```
maxdanilov, line 72: shall this workflow be in generaly having a timeout safety net that won't let it run forever? the risk I see now if that `gh run watch` can be stuck if the original workflow is taking too long to finish.

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+          gh run rerun ${{ inputs.run_id }} -R ${{ inputs.repository }} --failed
```
maxdanilov, line 73: is this operation waiting for the return workflow to finish, or just triggers a restart and does not wait for it to complete?

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
+name: Retrigger failed run action
```
maxdanilov, line 1: As this is a new action, it also needs a `README.md` to explain how it works and how to use it.

maxdanilov (changes requested):

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: The id of the run which we want to retrigger
+        required: true
+      error_message:
```
maxdanilov, line 9: what if a team wants to have multiple error message strings to match?

maxdanilov (changes requested):

rerun-failed-run/rerun-failed-run.yml
andromaqui: What about targeting this maybe in a follow up ticket as we would have to provide a way for them to see whether encountering all errors or at least one of the errors should retry the workflow?

maxdanilov: sure, sounds good. Just want to make sure the current architecture is not prohibiting that extension in the future.

rerun-failed-run/rerun-failed-run.yml
andromaqui: Good question, actually this wont wait until the workflow has finished it is just going to retrigger it. Looking at this again it is maybe redundant to watch the run

rerun-failed-run/rerun-failed-run.yml
andromaqui: yeah I agree, imo it is maybe actually not necessary to watch the run now that I think about it again - maybe it is enough to only retrigger it

rerun-failed-run/rerun-failed-run.yml
andromaqui: This is an example of a run where the log message has actually been encountered in the run:

Logs from the run start [here](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7830046910/job/21363225321#step:4:692) and finish [here](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7830046910/job/21363225321#step:4:1114). In these logs the inputted error message `Process completed with exit code 1.` is being encountered 3 times. Two of these times is because it is being inputted in the workflow [1](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7830046910/job/21363225321#step:4:1081), [2](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7830046910/job/21363225321#step:4:896) and one time because the error message is [actually encountered](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7830046910/job/21363225321#step:4:834). So if the error message is encountered in the logs then it will show up only two times.

andromaqui: [This](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7829741834/job/21362332769#step:4:1081) is an example from a workflow where the error message to compare against was `Random error message.`

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,73 @@
...
+        description: Custom error message to look for in the logs.
+        required: true
+      repository:
+        description: Github repository
```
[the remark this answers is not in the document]
andromaqui: Github actually doesnt allow placing : on the description neither escaping it. I will replace it with `GitHub repository (ORG/REPO_NAME)`

rerun-failed-run/rerun-failed-run.yml
maxdanilov: I see. Is there a way of excluding these 2 cases from count? The current implementation is quite brittle, relying that 2 extra mentions should happen in the logs. For example, if one day we stop passing `GH_DEBUG: api`  as an env var, it will be only printed once.

rerun-failed-run/rerun-failed-run.yml
andromaqui: I replaced it with  $GITHUB_OUTPUT

rerun-failed-run/rerun-failed-run.yml
andromaqui: This is actually a very valid point. I replaced it with 
`gh run view ${{ inputs.run_id }} -R ${{ inputs.repository }} --exit-status `

That command will show only the exit status of the run of the job and the reason why it fails

andromaqui: we can therefore search if there is more than 0 occurancies of the error message

rerun-failed-run/README.md
```diff
@@ -0,0 +1,70 @@
...
+Lastly, you can customize how many times you want the workflow to retrigger your job by configuring the number of attempts in the job level. See th example provided
+below for more information.
+
+## Inputs
+
+| Input name               | Description                                                                                   |
+|--------------------------|-----------------------------------------------------------------------------------------------|
+| run_id (required)        | The ID of the workflow run to retrigger.                                                      |
+| error_message (required) | Custom error message to look for in the logs                                                  |
+| repository (required)    | GitHub repository in the format ORG/REPO_NAME (example: `camunda/infra-global-github-action`) |
```
maxdanilov, line 32:
```suggestion
| repository (required)    | GitHub repository calling this reusable workflow, in the format ORG/REPO_NAME (example: `camunda/infra-global-github-action`) |
```

rerun-failed-run/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Workflow Example
+```yaml
+---
+ rerun-failed-jobs:
```
maxdanilov, line 37: I suggest rewriting this example as a full workflow, right now there's just `rerun-failed-jobs` defined. `job1`, `job2`, `job3` are mentioned, but it's not clear on what they are doing exactly.

I guess for the demo purposes of the readme, only `job1` would be sufficient, and it can be as simple as a job that just fails with a specific error message that we later look for in the `rerun-failed-jobs` example.

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,66 @@
...
+          # Fetch the logs from the previous run
+          gh config set pager cat
+          logs=$(gh run view ${{ inputs.run_id }} -R ${{ inputs.repository }} --exit-status || true)
+          echo "Logs coming from run with id: ${{ inputs.run_id }}"
+          echo "$logs"
+          error_count=$(echo "$logs" | grep -F "${{ inputs.error_message }}" | wc -l)
+          echo "error_count: $error_count"
+          if [ "$error_count" -gt 0 ]; then
+              echo "rerun=true" >>  $GITHUB_OUTPUT
+              echo "Error message ${{ inputs.error_message }} found more than two times in logs so the previous run has finished with that error."
```
maxdanilov, line 54: this error message needs to be updated (mentiones "more than two times")

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,66 @@
...
+          else
+              echo "rerun=false" >>  $GITHUB_OUTPUT
+              echo "Error message $error_message not found in logs"
```
maxdanilov, line 57: For better readability:
```suggestion
              echo "Error message '$error_message' not found in logs"
```

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,66 @@
...
+          echo "Fetching logs for run with id: ${{ inputs.run_id }} from ${{ inputs.repository }} repository."
+          # Fetch the logs from the previous run
+          gh config set pager cat
+          logs=$(gh run view ${{ inputs.run_id }} -R ${{ inputs.repository }} --exit-status || true)
+          echo "Logs coming from run with id: ${{ inputs.run_id }}"
+          echo "$logs"
+          error_count=$(echo "$logs" | grep -F "${{ inputs.error_message }}" | wc -l)
+          echo "error_count: $error_count"
+          if [ "$error_count" -gt 0 ]; then
+              echo "rerun=true" >>  $GITHUB_OUTPUT
```
maxdanilov, line 53: there's a trailing space:

```suggestion
              echo "rerun=true" >> $GITHUB_OUTPUT
```

rerun-failed-run/rerun-failed-run.yml
```diff
@@ -0,0 +1,66 @@
...
+              echo "Error message ${{ inputs.error_message }} found more than two times in logs so the previous run has finished with that error."
+          else
+              echo "rerun=false" >>  $GITHUB_OUTPUT
```
maxdanilov, line 56: there's a trailing space:
```suggestion
              echo "rerun=false" >> $GITHUB_OUTPUT
```

rerun-failed-run/README.md
```diff
@@ -0,0 +1,70 @@
+# rerun-failed-run
+
+This reusable workflow is aimed to target runs which have failed due to a specific error message and retrigger them. To do so, the workflow utilizes the
```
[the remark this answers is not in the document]
andromaqui: true this is much better

rerun-failed-run/README.md
```diff
@@ -0,0 +1,70 @@
...
+The workflow can be invoked with the aid of the github cli from another workflow. In case the calling workflow does not belong in the same repository a
+Github App Token should be generated in order to allow the invokation. For that purpose, a github app is already set up and the necessary `app id` and
+`app key` can be found in the following paths in Vault:
+
+```
+secret/data/products/infra/ci/retrigger-gha-workflow RETRIGGER_APP_KEY;
+secret/data/products/infra/ci/retrigger-gha-workflow RETRIGGER_APP_ID;
+```
+
+Additionally, in order to allow the workflow to be applied in a number of repositories and error messages, it is necessary to provide in the inputs not only
```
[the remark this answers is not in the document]
andromaqui: I think we can drop this then since it is kind of implied already that someone needs to specify these inputs

rerun-failed-run/README.md
andromaqui: makes sense, i made the `job1` just output a dummy string

rerun-failed-run/rerun-failed-run.yml
andromaqui: oh sorry I thought pre-commit would have caught that

rerun-failed-run/rerun-failed-run.yml
andromaqui: same as above, fixed it now

rerun-failed-run/rerun-failed-run.yml
andromaqui: I actually have the impression this is a bug and it should have been:
`Error message ${{ inputs.error_message } not found in logs
`

I changed it

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
+# rerun-failed-run
+
+This reusable workflow can be used to retrigger in GitHub Actions workflows to automatically retrigger them if they failed and contain a specific error message in their build logs.
```
maxdanilov, line 3:
```suggestion
This reusable workflow can be used in GitHub Actions workflows to automatically retrigger them if they failed and contain a specific error message in their build logs.
```

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Usage
+
+The workflow can be invoked with the aid of the GitHub CLI from another workflow. In case the calling workflow does not belong in the same repository the designated GitHub App Token should be generated in order to allow the invocation. For that purpose, a github app is already set up, and its credentials can be found in the following Vault secret paths:
```
maxdanilov, line 7:
```suggestion
The workflow can be invoked with the aid of the GitHub CLI from another workflow. In case the calling workflow is not in the same repository, the designated GitHub App Token should be generated to allow the invocation. For that purpose, a GitHub app is already set up, and its credentials can be found in the following Vault secret paths:
```

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+secret/data/products/infra/ci/retrigger-gha-workflow RETRIGGER_APP_KEY;
+secret/data/products/infra/ci/retrigger-gha-workflow RETRIGGER_APP_ID;
+```
+
+You can customize how many times you want the workflow to retrigger your job by configuring the number of attempts in the job level. See th example provided
```
maxdanilov, line 14:
```suggestion
You can specify how many times you want the workflow to retrigger your job by configuring the number of attempts at the job level. See the example provided
```

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
...
+below for more information.
+
```
maxdanilov, line 16: I'd also add a disclaimer that the retry limit must always be specified (otherwise there's a risk or re-running the original workflow forever)

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
...
+|--------------------------|-------------------------------------------------------------------------------------------------------------------------------|
+| run_id (required)        | The ID of the workflow run to retrigger.                                                                                      |
+| error_message (required) | Custom error message to look for in the logs                                                                                  |
+| repository (required)    | GitHub repository calling this reusable workflow, in the format ORG/REPO_NAME (example: `camunda/infra-global-github-action`) |
+
+### Workflow Example
+```yaml
+---
+jobs:
+ rerun:
```
maxdanilov, line 29: I think the name `rerun` is confusing here (not immediately clear whether it's the job to be rerun or the job to orchestrate a rerun of another job), can it be something more generic? like `job1` or similar?

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
...
+  runs-on: gcp-core-2-default
+  steps:
+    - name: Echo dummy text
+      run: echo "dummy text"
+
+ rerun-failed-jobs:
+    needs:
+      - job1
```
maxdanilov, line 37: above you're calling the job `rerun` but here it's `rerun`

rerun-failed-run/README.md
```diff
@@ -0,0 +1,66 @@
...
+    if: failure() && fromJSON(github.run_attempt) < 3
```
maxdanilov, line 38: let's add an explicit comment here that `fromJSON(github.run_attempt) < 3` is used a limitator to not retry the build more than 3 times (and is very important to have this condition)

maxdanilov (changes requested):
Adding small comments, mainly around documentation

rerun-failed-run/README.md
andromaqui: yeah this indeed makes sense, they should align

rerun-failed-run/README.md
andromaqui: yeah that is true

maxdanilov (approved):
LGTM, thank you!

PR #159, opened 2024-02-19
Move rerun-failed-run workflow to .github folder

andromaqui:
## Description
This pr moves the `rerun-failed-run.yml` to the folder `.github` as it contains a workflow_dispatch trigger. This sort of workflows need to be located in the `.github` folder as they otherwise cannot be run.

## Testing
Since I made a small change on the script as well, I made sure that the rerun mechanism still works when testing with web modelers repo some changes.
pr: https://github.com/camunda/web-modeler/pull/8087
workflow runs: [1](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7963868215) and [2](https://github.com/camunda/infra-test-retrying-failed-actions/actions/runs/7963875362)


rerun-failed-run/README.md
```diff
@@ -1,66 +0,0 @@
-# rerun-failed-run
```
maxdanilov, line 1: I don't want to lose this content here, I suggest making a `README.md` in `github/workflows` and putting it there. That readme file can be used for more documentation in the future, for the other workflows in that directory.

maxdanilov (changes requested):

maxdanilov (approved):

Issue #165, opened 2024-02-26
Enable multi-arch container image build (build-docker-image)


leiicamundi:
#### Problem
Our current GitHub Action (`docker/build-push-action@v5`) lacks multi-architecture support, limiting deployment to various production platforms.

#### Context
Referencing [Issue #40](https://github.com/camunda/team-infrastructure-experience/issues/40), our project infrastructure requires multi-arch images for broader platform compatibility.

#### Request
Enhance the action by integrating QEMU for multi-architecture support. Proposed workflow snippet:

```yaml
# ... (existing workflow)

    - name: Set up QEMU
      uses: docker/setup-qemu-action@v1

    - name: Set up Docker Buildx
      id: buildx
      uses: docker/setup-buildx-action@v1

    - name: Build and push Docker image
      id: build-push
      uses: docker/build-push-action@v5
      with:
        platforms: linux/amd64,linux/arm64,linux/arm/v7  # Add desired platforms

# ... (rest of the workflow)
```

#### Impact
Facilitating multi-arch builds improves deployment flexibility across diverse platforms.

PR #166, opened 2024-02-26
add multiarch docker image build

leiicamundi:
closes https://github.com/camunda/infra-global-github-actions/issues/165

### DIFF
--- a/build-docker-image/action.yml
+++ b/build-docker-image/action.yml
@@ -50,7 +50,7 @@ inputs:
   qemu_image:
     description: QEMU static binaries Docker image
     required: false
-    default: 'tonistiigi/binfmt:latest'
+    default: 'tonistiigi/binfmt:master'
   qemu_platforms:
     description: Platforms to install (e.g., `arm64,riscv64,arm`)
     required: false


build-docker-image/action.yml
```diff
@@ -43,6 +47,14 @@ inputs:
     description: Whether load docker image from build into local Docker Images
     required: false
     default: 'false'
+  qemu_image:
+    description: QEMU static binaries Docker image
+    required: false
+    default: 'tonistiigi/binfmt:latest'
```
Langleu, line 53: you'd probably want to use `tonistiigi/binfmt:master`, `latest` has been outdated for ~2 years.

Langleu (approved):
besides the comment it looks good 👍.
Did some checks with a repo that is using this action and it's
- [x] backwards compatible
- [x] builds and pushes multiple archs

build-docker-image/action.yml
Langleu: I'd go with `master` for now.
See also https://github.com/docker/setup-qemu-action?tab=readme-ov-file#about where they are using `master` instead of `latest` even though every piece of their own documentation points to `latest`.

Langleu (approved):

Issue #167, opened 2024-03-01
Python Installation Failure on ARM Architecture


leiicamundi:
### Problem
Encountered a Python installation failure when using AWS ARM runners for self-hosted workflows.

### Description
While configuring self-hosted runners, I opted to use AWS ARM runners scheduled on arm64 nodes.

However, during the setup process, I encountered a Python installation failure with the following error:

```
The version '3.12' with architecture 'arm64' was not found for Ubuntu 20.04.
```

### Solution
The recommended solution is to use deadsnakes for ARM, as demonstrated in the following workflow snippet:

```yaml
- name: Install Python3 (ubuntu-arm install)
  uses: deadsnakes/action@3.1.0
  # actions/setup-python doesn't yet support ARM, see https://github.com/actions/setup-python/issues/108#issuecomment-1756948951
  if: ${{ matrix.os == 'ubuntu-arm' }}
```

**Note:** This solution is currently the only available option until official ARM support for Linux is provided by `actions/setup-python`. Refer to [Issue #108](https://github.com/actions/setup-python/issues/108) for ongoing discussions on ARM support.

### Impact
This Python installation issue is particularly noticeable when using AWS ARM runners for self-hosted workflows, we also need to test it on arm runners.

PR #168, opened 2024-03-01
Feature/arm python

leiicamundi:
closes https://github.com/camunda/infra-global-github-actions/issues/167

### DIFF
--- a/common-tooling/action.yml
+++ b/common-tooling/action.yml
@@ -120,7 +120,7 @@ inputs:
     default: 'true'
   # Python
   python-version:
-    description: "Version range or exact version of Python or PyPy to use, using SemVer's version range syntax. Reads from .python-version if unset."
+    description: "Version range or exact version of Python or PyPy to use, specified using SemVer's version range syntax. If unset, it reads from .python-version. Please note that on ARM, the list of available versions differs, and you can find the list of Python distributions at https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa/+packages."
     default: "3.11"
   python-version-file:
     description: "File containing the Python version to use. Example: .python-version (not supported on arm)"
--- a/.github/workflows/test-common-tooling.yml
+++ b/.github/workflows/test-common-tooling.yml
@@ -89,5 +89,5 @@ jobs:
         node-version: 20 # see https://github.com/actions/setup-node#supported-version-syntax for supported syntax
         java-version: 20 # see https://github.com/marketplace/actions/setup-java-jdk#supported-version-syntax for supported syntax
         java-distribution: adopt # see https://github.com/marketplace/actions/setup-java-jdk#supported-distributions for supported distros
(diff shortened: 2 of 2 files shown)


common-tooling/action.yml
```diff
@@ -262,6 +264,32 @@ runs:
...
+    with:
+      python-version: ${{ inputs.python-version }}
+      python-version-file: ${{ inputs.python-version-file }}
+      cache: ${{ inputs.python-cache }}
+      architecture: ${{ inputs.python-architecture }}
+      check-latest: ${{ inputs.python-check-latest }}
+      token: ${{ inputs.python-token }}
+      cache-dependency-path: ${{ inputs.python-cache-dependency-path }}
+      update-environment: ${{ inputs.python-update-environment }}
+      allow-prereleases: ${{ inputs.python-allow-prereleases }}
```
Langleu, line 281: the only inputs supported by `deadsnakes/action` are
- python-version
- debug
- nogil

common-tooling/action.yml
```diff
@@ -262,6 +264,32 @@ runs:
       cache-dependency-path: ${{ inputs.python-cache-dependency-path }}
       update-environment: ${{ inputs.python-update-environment }}
       allow-prereleases: ${{ inputs.python-allow-prereleases }}
+
+  - name: Install Python3 (ubuntu-arm install)
+    uses: deadsnakes/action@v3.1.0
+    # actions/setup-python doesn't yet support ARM, see https://github.com/actions/setup-python/issues/705#issuecomment-1756948951
+    if: ${{ contains(runner.arch, 'arm') && contains(runner.name, 'ubuntu') && inputs.python-enabled == 'true' && ( env.python_missing == 'true' || inputs.overwrite == 'true' ) }}
```
Langleu, line 271: I don't think this is executed on arm runners since the if can never be true due to `contains(runner.name, 'ubuntu'`) afaik self-hosted runners should never contain `ubuntu` in their name, therefore skipped.
I also didn't find it called in the logs.

common-tooling/action.yml
```diff
@@ -262,6 +264,32 @@ runs:
...
+      cache: ${{ inputs.python-cache }}
+      architecture: ${{ inputs.python-architecture }}
+      check-latest: ${{ inputs.python-check-latest }}
+      token: ${{ inputs.python-token }}
+      cache-dependency-path: ${{ inputs.python-cache-dependency-path }}
+      update-environment: ${{ inputs.python-update-environment }}
+      allow-prereleases: ${{ inputs.python-allow-prereleases }}
+
+  - name: Add Python3 Shortcut and add pip (ARM)
+    if: ${{ contains(runner.arch, 'arm') }}
```
Langleu, line 284: I'd suggest to align this with the previous if so it's triggered under the same conditions.

Langleu (approved):
nice work 🚀

PR #220, opened 2024-06-18
feat: on workflow retry, wait for completion

leiicamundi:
This PR introduces a mechanism to wait for the workflow that needs to be retried to complete properly. Without this waiting mechanism, the retry fails: [workflow link](https://github.com/camunda/infra-global-github-actions/actions/runs/9568479702/job/26378796649).

I haven't been able to test it since the action calls a workflow_dispatch.

This scenario occurs when cleanup actions exist after capturing the fail handover (https://github.com/camunda/keycloak/pull/131).

### DIFF
--- a/.github/workflows/rerun-failed-run.yml
+++ b/.github/workflows/rerun-failed-run.yml
@@ -58,4 +58,6 @@ jobs:
           GH_DEBUG: api
           GITHUB_TOKEN: ${{ steps.github-token.outputs.token }}
         run: |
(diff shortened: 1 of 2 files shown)


cmur2 (approved):
Thanks for discovering this edge case! The suggested fix makes sense to avoid this edge case while not impacting any other legitimate usage as far as I can see + nor increasing cost  (due to increased runtime) unnecessarily.

PR #228, opened 2024-07-01
Action to comment on PR with current deployment results

kellervater:
As described in [this issue](https://github.com/camunda/team-infrastructure/issues/459) we want to enhance the ux by summarizing deployment results of preview environments in a comment on the PR.

This PR adds 4 basic things:
* a new action `preview-env/comment`, which is capable of upserting a summarize comment in the PR by grabbing all deployment artifacts of the same workflow run attempt the action is being called
* Documentation of the action (purpose, usage)
* an adaption of the `preview-env/create` action, which creates and uploads said deployment artifacts. The artifacts themselve are separate markdown files for each possible outcome (success/failure/cancelled)
* Removes the comment when being called without any uploaded deployment artifacts

After a successfull review I'll intend to squash all commits together, before merging.


preview-env/comment/action.yml
```diff
@@ -0,0 +1,39 @@
+---
+name: Create Deployment Result Summary
+description: |
+  Creates a deployment result summary as comment in the current PR when deployment deployment artifacts have been uploaded during this run.
```
clementnero, line 4: The aim of this composite action is to manage comments (create, update, and probably deletion ?), so the name should reflect it.

```suggestion
name: Manage Deployment Result Summary
description: |
  Manages a deployment result summary as comment in the current PR
```

preview-env/comment/action.yml
```diff
@@ -0,0 +1,39 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Download Deployment Status Artifacts
+      uses: actions/download-artifact@v4
+      with:
+        path: deployment_status
+        pattern: deployment-status-${{ github.run_id }}-${{ github.run_attempt }}-*
+
```
clementnero, line 14: Blank new line to be removed?

preview-env/comment/action.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Check for deployment artifacts
+      shell: bash
+      run: |
+        if [ -d deployment_status ] && find deployment_status -type f -name '*.md' | grep -q .; then
+          echo "artifacts_found=true" >> $GITHUB_ENV
+        else
+          echo "artifacts_found=false" >> $GITHUB_ENV
+        fi
```
clementnero, line 22: I would suggest to use proper outputs (`GITHUB_OUTPUT`) ? as the env variables defined there is never used as is.

It's too bad the actions/download-artifact@v4 action does not provide proper outputs :(

preview-env/comment/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      if: env.artifacts_found == 'true'
+      shell: bash
+      run: |
+        summary="# :rocket: Deployment Results :rocket:\n"
+        for file in deployment_status/**/*.md; do
+          result=$(cat "$file")
+          summary+=$(cat "$file")
+        done
+        echo -e "$summary" > summary.md
+        echo -e "$summary"
```
clementnero, line 33: you can use the `tee` command to merge these two lines?

preview-env/comment/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        fi
+    - name: Create deployment status summary
+      if: env.artifacts_found == 'true'
+      shell: bash
+      run: |
+        summary="# :rocket: Deployment Results :rocket:\n"
+        for file in deployment_status/**/*.md; do
+          result=$(cat "$file")
+          summary+=$(cat "$file")
+        done
```
clementnero, line 31: I think you can simplify it with a unique `cat` command ?

```suggestion
summary+=$(cat deployment_status/**/*.md)
```

I assume the `result` variable is a leftover ?

preview-env/create/action.yml
```diff
@@ -105,11 +105,11 @@ runs:
       ARGOCD_AUTH_TOKEN: ${{ inputs.argocd_token }}
       ARGOCD_OPTS: "--grpc-web"
       ARGOCD_SERVER: ${{ inputs.argocd_server }}
-  - name: Wait for sync (up to 20 minutes)
+  - name: Wait for sync (up to 2 minutes)
```
clementnero, line 108: I assume this is a leftover, used for testing, which must now must be removed ... ?

preview-env/create/action.yml
```diff
@@ -105,11 +105,11 @@ runs:
...
     shell: bash
     run: |-
       echo "::group::wait for sync"
-      argocd app wait ${{ inputs.app_name }} --timeout 1200 --health
+      argocd app wait ${{ inputs.app_name }} --timeout 120 --health
```
clementnero, line 112: Same as above ?

preview-env/create/action.yml
```diff
@@ -133,3 +133,19 @@ runs:
...
       env: ${{ inputs.app_name }}
+  - name: Create deployment status artifact
+    if: always()
+    shell: bash
+    run: |
+      if [[ "${{ github.action_status }}" == "success" ]]; then
+        echo ":full_moon_with_face: [\`${{ inputs.app_name }}\`](${{ inputs.app_url }}): ${{ github.action_status }}\n" > status_${{ inputs.app_name }}.md
+      else
+        echo ":boom: \`${{ inputs.app_name }}\`: ${{ github.action_status }} :arrow_right: see [ArgoCD](https://${{ inputs.argocd_server }}/applications?health=Progressing%2CDegraded&search=${{ inputs.app_name }})\n" > status_${{ inputs.app_name }}.md
+      fi
```
clementnero, line 144: I think you can make this step more readable by defining the two different templates in separate env variables at the action level, and use YAML multi-line capabilities ... 
Same for the `github.action_status` value, as from a security point of view it is recommended to avoid injecting context values directly into the script, and to use environment variables instead.

```
  - name: Create deployment status artifact
    if: always()
    shell: bash
    env:
      TEMPLATE_A: >
        my
        template A
      TEMPLATE_B: |
        my template B
      STATUS: ${{ github.action_status }}
    run: |
      if [[ "$STATUS" == "success" ]]; then
        echo "$TEMPLATE_A" > status_${{ inputs.app_name }}.md
      else
        echo "$TEMPLATE_B" > status_${{ inputs.app_name }}.md
      fi
```

preview-env/create/action.yml
```diff
@@ -133,3 +133,19 @@ runs:
...
+  - name: Upload deployment status as artifact
+    if: always()
+    uses: actions/upload-artifact@v4
+    with:
+      name: deployment-status-${{ github.run_id }}-${{ github.run_attempt }}-${{ inputs.app_name }}
+      path: status_${{ inputs.app_name }}.md
+      retention-days: 1
```
clementnero, line 151: If there is no need to keep the artifact once the workflow is finished (e.g. for rerun or so), we probably can remove it ? Perhaps we can mark it as ephemeral ?

clementnero (changes requested):
Thx for coming up with this solution based on temporary artifacts to smoothly collect results. It's seems to work fine.
I left a couple of remarks directly in the code.

If the deployment is removed, it makes sense to me to remove the comment as well ? What do you think ? In this case, if there is no artifact, the action can just remove the comment ?

kellervater:
> If the deployment is removed, it makes sense to me to remove the comment as well ? What do you think ? In this case, if there is no artifact, the action can just remove the comment ?

@clementnero 
I would then do the following: 
* modify the `comment` action in a way, that it's removing the comment, if it wasn't able to any artifacts
* call the `comment` action at the end of the `preview-env/cleanup` action, to make sure the comment gets removed when the preview-env is being removed

clementnero:
> call the comment action at the end of the preview-env/cleanup action, to make sure the comment gets removed when the preview-env is being removed

You mean `preview-env/destroy` ?

That makes total sense. Additionally, now the composite action is used twice, which justifies its existence ;)

preview-env/create/action.yml
kellervater: unfortunately, there's no built-in option to mark it as ephemeral. 
That's the reason why I chose the minimum retention value.
So I judged, that it doesn't have a big footprint, hence it shouldn't hurt to keep it for a day.

kellervater:
agreed and reworked all suggestions as well as the newly aligned "delete" functionality.

Also improved a bit on the summary function.

## Summaries
### PR
Possible summaries now look like this:
image
See [this comment](https://github.com/camunda/web-modeler/pull/10038#issuecomment-2248232941) for details.

### Workflow Summary
It was a small step for a man but a huge leap Github-kind....

... to add each summary entry to the Workflow's summary page as well. It looks like this:
image
You can find it [here](https://github.com/camunda/web-modeler/actions/runs/9761046876?pr=10038)

## Deletion functionality
As mutually agreed, it does make sense to also remove the PR's summary comment, when the environments got teared down.
Here I added a behavior to the comment action, which deletes the comment, when being called without any artifacts uploaded following the naming pattern `deployment-status-${{ github.run_id }}-${{ github.run_attempt }}-*`

To test this behavior, I used the `optimize` Draft PR:
https://github.com/camunda/camunda-optimize/pull/13703
Since the comment got deleted, there's not much to see on the PR itself, but the `label removed` message.

To show the deletion I attached the deletion logs here: https://github.com/camunda/camunda-optimize/actions/runs/10079366095/job/27866521743?pr=13703#step:5:249

preview-env/comment/action.yml
```diff
@@ -0,0 +1,46 @@
...
+      uses: mshick/add-pr-comment@v2
+      with:
+        message-path: summary.md
+        refresh-message-position: true
+    - name: Delete Summary
+      # this action will cause a warning:
+      # `Unexpected input(s) 'owner', 'repo', 'issue_number', valid inputs are ['route', 'mediaType']`
+      # This is expected and further info can be found here: https://github.com/octokit/request-action?tab=readme-ov-file#warnings
+      if: steps.check_artifacts.outputs.ARTIFACTS_FOUND == 'true'
+      uses: aki77/delete-pr-comments-action@v2
```
clementnero, line 44: As discussed, it's indeed better to use a GHA than custom implementation here 👍

preview-env/create/action.yml
```diff
@@ -133,3 +133,24 @@ runs:
       env_url: ${{ inputs.app_url }}
       ref: ${{ inputs.revision }}
       env: ${{ inputs.app_name }}
+  - name: Create deployment status artifact
+    if: always()
+    env:
+      TEMPLATE_SUCCESS: '__SUCCESS__ :white_check_mark: [ArgoCD](https://${{ inputs.argocd_server }}/applications?search=${{ inputs.app_name }}) :link: [`${{ inputs.app_name }}`](${{ inputs.app_url }})'
```
clementnero, line 139: I like it! You can even split the long string in multiple lines ;)

preview-env/destroy/action.yml
```diff
@@ -55,6 +55,9 @@ runs:
       ARGOCD_OPTS: "--grpc-web"
       ARGOCD_SERVER: ${{ inputs.argocd_server  }}
 
+  - name: Summarize Deployment Results
```
clementnero, line 58: As agreed, this step will be moved to the teardown workflows for consistency.

clementnero (changes requested):
Thx for taking into account my feedbacks and discussing the last remarks. I just left some notes as reminder.

preview-env/comment/action.yml
kellervater: as discussed afterwards, went with the shell-script approach

preview-env/create/action.yml
kellervater: created separate files for the templates

kellervater:
Hey @clementnero

## Next Deletion Approach 
As requested: I changed the deletion methods from self-written to the Github action [`delete-pr-review-comments`](https://github.com/marketplace/actions/delete-pr-review-comments).
This didn't work out and so I changed (like we've discussed) the approach yet once more to make use of already existing shell scripts as well as another wrapper around it.

The deletion is working fine using bash scripts.

Here's the showcase PR for the deletion: https://github.com/camunda/camunda-optimize/pull/13703
Example Logs of the deletion itself can be found here: https://github.com/camunda/camunda-optimize/actions/runs/10100061158/job/27930596790?pr=13703

## Changed format of summary
Like you can see via [this PR](https://github.com/camunda/web-modeler/pull/10038#issuecomment-2251242381) the format should look like we've discussed in Slack:
image

@clementnero 
kindly asking for the ReReReview ;-)

preview-env/comment/README.md
```diff
@@ -0,0 +1,155 @@
...
+        echo ":boom: \`${{ inputs.app_name }}\`: ${{ github.action_status }} :arrow_right: see [ArgoCD](https://${{ inputs.argocd_server }}/applications?health=Progressing%2CDegraded&search=${{ inputs.app_name }})\n" > status_${{ inputs.app_name }}.md
+      fi
+  - name: Upload deployment status as artifact
+    if: always()
+    uses: actions/upload-artifact@v4
+    with:
+      name: deployment-status-${{ github.run_id }}-${{ github.run_attempt }}-${{ inputs.app_name }}
+      path: status_${{ inputs.app_name }}.md
+      retention-days: 1
+```
```
clementnero, line 152: This part is not up to date, due to the recent changes. I would replace it by a single comment and/or link.

preview-env/comment/action.yml
```diff
@@ -0,0 +1,51 @@
...
+      if: steps.check_artifacts.outputs.ARTIFACTS_FOUND == 'true'
+      shell: bash
+      run: |
+        shopt -s globstar   # enables the `**` wildcard
+        summary="## :rocket: Deployment Results :rocket:\n"
+        summary+="Please find below the results of the latest deployments.\n"
+        summary+=$(cat deployment_status/**/*.md)
+        summary+="\n## :wrench: Troubleshooting :wrench:\n"
+        summary+="The :link: \`ArgoCD\` link can be used to check the state and configuration of all the services deployed as part of the preview environments and get logs.\n\n"
+        summary+="In case of error, please check :clipboard: \`Deployment Jobs\` and :link: \`ArgoCD\` to debug, and check our (yet to come) troubleshooting page"
```
clementnero, line 35: Could we use a template file(s) here (as you did for the `create` composite action) ? It would be better for maintenance and readability.

clementnero: I just saw the following for the `message-path` attribute: https://github.com/mshick/add-pr-comment?tab=readme-ov-file#configuration-options

> Supports multi-line input and globs. Multiple messages will be concatenated.

Meaning you can eventually simplify all this with something like

```yaml
message-path: |
  templates/head.md
  deployment_status/**/*.md
  templates/troubleshooting.md
```

preview-env/comment/action.yml
```diff
@@ -0,0 +1,51 @@
...
+        echo -e "$summary" | tee summary.md
+    - name: Upsert Deployment Result Summary
+      if: steps.check_artifacts.outputs.ARTIFACTS_FOUND == 'true'
+      uses: mshick/add-pr-comment@v2
+      with:
+        message-path: summary.md
+        refresh-message-position: true
+    - name: Delete Summary
```
clementnero, line 43: To be consistent with the upsert step, I would rename it to 
```suggestion
    - name: Delete Deployment Result Summary
```

preview-env/comment/action.yml
```diff
@@ -0,0 +1,51 @@
...
+      if: steps.check_artifacts.outputs.ARTIFACTS_FOUND == 'false'
+      env:
+        COMMENT_BODY: '<!-- add-pr-comment:add-pr-comment -->'
```
clementnero, line 46: Can we adjust the name of the env var to make it more clear ?
```suggestion
        TAG_IN_COMMENT_BODY: '<!-- add-pr-comment:add-pr-comment -->'
```

preview-env/comment/delete.sh
```diff
@@ -0,0 +1,47 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+RELATIVE_SCRIPT_PATH=$(dirname "$0")
+COMMENT_BODY=${COMMENT_BODY?}
+PULL_REQUEST=${PULL_REQUEST?}
+
+# shellcheck source=/dev/null
+source "$RELATIVE_SCRIPT_PATH/../clean/github.sh"
```
clementnero, line 19: I was wondering if this would work 👍 
It's great, because we can export it to a more general folder next time.

preview-env/create/action.yml
```diff
@@ -133,3 +133,23 @@ runs:
...
+    if: always()
+    shell: bash
+    env:
+      APP_NAME: ${{ inputs.app_name }}
+      APP_URL: ${{ inputs.app_url }}
+      ARGOCD_SERVER: ${{ inputs.argocd_server }}
+      REPO: ${{ github.repository }}
+      RUN_ID: ${{ github.run_id }}
+    run: |
+      template=$(envsubst < ${{ github.action_path }}/comment_template_${{ github.action_status }}.md)
```
clementnero, line 146: I would put the templates in a dedicated folder like:
- templates/comment-failure.md
- templates/comment-success.md
- templates/comment-cancelled.md

preview-env/comment/action.yml
```diff
@@ -0,0 +1,51 @@
...
+  The deployment artifact name pattern: `deployment-status-<runID>-<runAttempt>-*`
+runs:
+  using: composite
+  steps:
+    - name: Download Deployment Status Artifacts
+      uses: actions/download-artifact@v4
+      with:
+        path: deployment_status
+        pattern: deployment-status-${{ github.run_id }}-${{ github.run_attempt }}-*
+    - name: Check for deployment artifacts
```
clementnero, line 16: To be consistent with the step name above
```suggestion
    - name: Check for Deployment Status Artifacts
```

preview-env/comment/action.yml
```diff
@@ -0,0 +1,51 @@
...
+        summary+="In case of error, please check :clipboard: \`Deployment Jobs\` and :link: \`ArgoCD\` to debug, and check our (yet to come) troubleshooting page"
+        echo -e "$summary" | tee summary.md
+    - name: Upsert Deployment Result Summary
+      if: steps.check_artifacts.outputs.ARTIFACTS_FOUND == 'true'
+      uses: mshick/add-pr-comment@v2
+      with:
+        message-path: summary.md
+        refresh-message-position: true
+    - name: Delete Summary
+      if: steps.check_artifacts.outputs.ARTIFACTS_FOUND == 'false'
```
clementnero, line 44: It's more a personal remark (as part of this peering exercise), but I prefer keeping outputs in lower case:
```suggestion
      if: steps.deployment-status-artifacts.outputs.found == 'false'
```

Up to you to change if you think it makes more sense.

clementnero (changes requested):
Thank you again for taking into account my remarks. The implementation of the new action is now crisp and clean, and reuses existing functionality. It's really good 🚀 

I have left a couple of remarks for minor details still to be addressed.

preview-env/comment/delete.sh
kellervater: haha... my thought exactly. Since I wasn't sure if it works at all, i just "crapped" it here like it is now.
But I had the same idea of extracting this in a kind of "util" folder.

preview-env/comment/action.yml
kellervater: I'll comply with your suggestion since 
* i have no strong opinion on this 
* I like to have a common approach with fellow teammates

kellervater:
## Rework
Done! Every point should be addressed by now.

Tested deletion in a little modified workflow [in here](https://github.com/camunda/web-modeler/actions/runs/9761046876/job/27961903357) (toggled deletion flag for the test)

Tested comment summary can be found [here](https://github.com/camunda/web-modeler/pull/10038#issuecomment-2252636369)!

@clementnero pls review

preview-env/comment/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Possible Deployment Results
+All of the following deployment results will be collected in a comment with a heading like this:
+
+> # :rocket: Deployment Results :rocket:
+
+For `preview-env-create` we have 3 different deployment result templates:
+
+### Success
+Please see [this template](../create/templates/comment_template_success.md).
```
clementnero, line 17: Wrong path ;)

clementnero, line 17: Wrong path to the template ;)

clementnero (approved):
It's now look very good 🚀 

I noted one typo but not blocking.

PR #237, opened 2024-07-30
ci(preview-env): fix argoCD template

kellervater:
This PR is a fix for the comment summary templates as aligned in:
* https://github.com/camunda/team-infrastructure/issues/459#issuecomment-2258008062

It specifically fixes the ArgoCD URLs in the result templates.

Tested through this Draft PR: <tbd>

### DIFF
--- a/preview-env/create/templates/comment-cancelled.md
+++ b/preview-env/create/templates/comment-cancelled.md
@@ -1,4 +1,4 @@
 ### `${APP_NAME}`
 * __Status:__ :x:
-* __ArgoCD:__ :link: [Link](https://${ARGOCD_SERVER}/applications?search=${APP_NAME})
+* __ArgoCD:__ :link: [Link](https://${ARGOCD_SERVER}/applications/argocd/${APP_NAME}?view=tree&resource=)([degraded resources](https://${ARGOCD_SERVER}/applications/argocd/${APP_NAME}?view=tree&resource=health%3ADegraded))
 * __Deployment Jobs:__ :clipboard: [Link](https://github.com/${REPO}/actions/runs/${RUN_ID})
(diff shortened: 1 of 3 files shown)


kellervater:
The new behavior on failed deployments:
image

They now lead directly to the resources(degraded resources) as requested.
[Link](https://argocd.int.camunda.com/applications/argocd/web-modeler-gh-459ment-summary-fix-saas?view=tree&resource=)([degraded resources](https://argocd.int.camunda.com/applications/argocd/web-modeler-gh-459ment-summary-fix-saas?view=tree&resource=health%3ADegraded))

The same applies for for successful deployments (without the degradation link):
image
The link now looks like this:
https://argocd.int.camunda.com/applications/argocd/web-modeler-gh-459ment-summary-fix-saas?view=tree&resource=

In this special case, we had a timing issue during the first try, which resolved itself after sometime, since the build couldn't keep up with the deployment.
Thus I added another section in the footer to outline, that the actual truth is always to be found in ArgoCD like so:

> [!NOTE]
> There's a chance, that timing issues cause a deployment to be shown as failure here, but are actually still progressing or even became healthy in the meantime in ArgoCD. When in doubt: Just check both ArgoCD links!

preview-env/comment/templates/comment-footer.md
```diff
@@ -2,3 +2,6 @@
 The :link: `ArgoCD` link can be used to check the state and configuration of all the services deployed as part of the preview environments and get logs.
 
 In case of error, please check :clipboard: `Deployment Jobs` and :link: `ArgoCD` to debug, and check our (yet to come) troubleshooting page
+
+> [!NOTE]
+> There's a chance, that timing issues cause a deployment to be shown as failure here, but are actually still progressing or even became healthy in the meantime in ArgoCD. When in doubt: Just check both ArgoCD links!
```
clementnero, line 7: I think this note is a bit overwhelming for users at this point. I would remove it.

clementnero (approved):
Only one remark regarding the new note, but nothing blocking.
Looks good to me, thank you!!  🚀

preview-env/comment/templates/comment-footer.md
clementnero: especially when all deployment are successful.

kellervater: agree. Glad I put it into a separate commit. Removed it now :-)

PR #255, opened 2024-09-04
feat(setup-yarn-cache): add new action, from the C8 monorepo

cmur2:
This action was originally created in https://github.com/camunda/camunda/issues/21427 but is now relocated for wider use also in other repositories (disccused in [Slack](https://camunda.slack.com/archives/CHY2S7KDJ/p1725364793383679?thread_ts=1725281533.019179&cid=CHY2S7KDJ)).

Noteworthy is the new input for this action `cache_create_branch_regex` which allows configuring on which branches to save the cache on (done via Bash now instead of GHA expression to allow parameterization), the defaults suit the C8 monorepo to avoid repeating configuration.

The new action is successfully used in https://github.com/camunda/camunda/actions/runs/10699692498/job/29661839825?pr=21915 as an example


kellervater (approved):
Looking great.

Just one question on this:
Do you think it's worth observing [this setup-node issue](https://github.com/actions/setup-node/issues/663)? If they come up with a built-in solution it could make sense.

cmur2:
> Do you think it's worth observing [this setup-node issue](https://github.com/actions/setup-node/issues/663)? If they come up with a built-in solution it could make sense.

I subscribed to it now but given the timeline I'm not expecting much soon.

PR #271, opened 2024-09-30
feat: use official action for gh token generation

leiicamundi:
GitHub is now maintaining an official GHA for creating app tokens: [actions/create-github-app-token](https://github.com/actions/create-github-app-token), which is intended to replace [tibdex/github-app-token](https://github.com/tibdex/github-app-token).

The official GHA generates a token that, by default, remains valid until the end of the job, whereas the previous GHA created a non-configurable one-hour token. I also took the opportunity to update the outputs and add an option for generating non-expiring tokens.

The token generation has been tested in this run: [camunda-tf-eks-module GitHub Actions Run](https://github.com/camunda/camunda-tf-eks-module/actions/runs/11107455213/job/30858074950?pr=130).

The retry action has been tested in https://github.com/camunda/keycloak/actions/runs/11108559099/job/30861837294?pr=190, the associated workflow works correctly (https://github.com/camunda/infra-global-github-actions/actions/runs/11108577107/job/30861848357). For this one, I had to fix the retrieval of the logs, it was not working correctly (empty output) without the `--log-failed` flag.

### DIFF
--- a/.github/workflows/rerun-failed-run.yml
+++ b/.github/workflows/rerun-failed-run.yml
@@ -71,8 +71,6 @@ jobs:
           gh config set pager cat
           logs=$(gh run view ${{ inputs.run_id }} -R ${{ inputs.repository }} --exit-status --log-failed || true)
           echo "Logs coming from run with id: ${{ inputs.run_id }}"
-          # TODO: remove before merge
-          echo "Debug logs: $logs"
           inputted_errors='${{ inputs.error_messages }}'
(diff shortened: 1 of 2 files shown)


kellervater (approved):
LGTM, but one question about

> I also took the opportunity to update the outputs and add an option for generating non-expiring tokens.
Out of interest: Does that mean that these tokens never expire at all or is there some default lifetime?

PR #277, opened 2024-10-07
feat: implements owner and repositories input for token generation

leiicamundi:
This PR adds the `owner` and `repositories` parameters inherited from the official GitHub action.

By default, the generated tokens have a scope limited to the current repository. To extend this scope to other repositories, you can list them in the `repositories` field. If you want the scope to apply to all repositories of the owner, you should specify the value `!null`.

I have tested this in the following workflows:
- [Run 11220366350](https://github.com/camunda/team-infrastructure-experience/actions/runs/11220366350/job/31188374915)
- [Run 11220233216](https://github.com/camunda/team-infrastructure-experience/actions/runs/11220233216/job/31187940158)

### DIFF
--- a/generate-github-app-token-from-vault-secrets/README.md
+++ b/generate-github-app-token-from-vault-secrets/README.md
@@ -19,7 +19,7 @@ This composite GHA can be used in any repository.
 | vault-url                           | The URL for the Vault endpoint                                          |
 | skip-token-revoke                   | If truthy, the token will not be revoked when the current job is complete (optional) |
 | owner                               | The owner of the GitHub App installation (defaults to current repository owner, optional). |
-| repositories                        | Comma or newline-separated list of repositories to install the GitHub App on (defaults to current repository if owner is unset, optional). If you want to generate a token that has access to all repositories of the owner, set this to `!all` and explicitely set an `owner`. |
+| repositories                        | Comma or newline-separated list of repositories for which the GitHub app token will be valid for(defaults to current repository if owner is unset, optional). If you want to generate a token that has access to all repositories of the owner, set this to `!all` and explicitely set an `owner`. |
 
 > (*) The above Vault properties only support App Role authentication (`vault-auth-method=approle`) for now. New inputs may be added in the future to support other authentication methods.
 


generate-github-app-token-from-vault-secrets/README.md
```diff
@@ -18,6 +18,8 @@ This composite GHA can be used in any repository.
 | vault-auth-secret-id                | The Secret Id for (Vault) App Role authentication                       |
 | vault-url                           | The URL for the Vault endpoint                                          |
 | skip-token-revoke                   | If truthy, the token will not be revoked when the current job is complete (optional) |
+| owner                               | The owner of the GitHub App installation (defaults to current repository owner, optional). To leave this value empty in the token generation, set it to `!null`. |
```
maxdanilov, line 21: why not just empty string "" instead? I think it's more intuitive like that.

generate-github-app-token-from-vault-secrets/README.md
```diff
@@ -18,6 +18,8 @@ This composite GHA can be used in any repository.
...
+| repositories                        | Comma or newline-separated list of repositories to install the GitHub App on (defaults to current repository if owner is unset, optional). If you want to generate a token that has access to all repositories of the owner, set this to `!null`. |
```
maxdanilov, line 22: same as above

maxdanilov (changes requested):

generate-github-app-token-from-vault-secrets/README.md
maxdanilov: I'd also argue that there's the need to have the case of generating a token for all repositories - normally it's either the current one, or a different (specific) one, wdyt?

generate-github-app-token-from-vault-secrets/README.md
```diff
@@ -18,6 +18,8 @@ This composite GHA can be used in any repository.
 | vault-auth-secret-id                | The Secret Id for (Vault) App Role authentication                       |
 | vault-url                           | The URL for the Vault endpoint                                          |
 | skip-token-revoke                   | If truthy, the token will not be revoked when the current job is complete (optional) |
+| owner                               | The owner of the GitHub App installation (defaults to current repository owner, optional). |
+| repositories                        | Comma or newline-separated list of repositories to install the GitHub App on (defaults to current repository if owner is unset, optional). If you want to generate a token that has access to all repositories of the owner, set this to `!all` and explicitely set an `owner`. |
```
maxdanilov, line 22: since the app is not installed by this action, only a token obtained:
```suggestion
| repositories                        | Comma or newline-separated list of repositories for which the GitHub app token will be valid for(defaults to current repository if owner is unset, optional). If you want to generate a token that has access to all repositories of the owner, set this to `!all` and explicitely set an `owner`. |
```

generate-github-app-token-from-vault-secrets/README.md
maxdanilov: basically, wouldn't it be enough to have a thin wrapper for the https://github.com/actions/create-github-app-token and just to pass `owner` and `repositories` to it as is?

maxdanilov (commented):
Leaving some follow-up comments after thinking about it.

generate-github-app-token-from-vault-secrets/README.md
maxdanilov: Thank you, I didn't think of the Renovate use case, it makes total sense.

maxdanilov (approved):
Thank you for the improvement and addressing the changes, LGTM!

PR #279, opened 2024-10-17
fix(download-center): setup python with supported version

barmac:
This is required due to gsutil not supporting Python 3.12 and newer. Cf. https://github.com/camunda/camunda-modeler/actions/runs/11371346205/job/31633203620#step:9:95

### DIFF
--- a/download-center-upload/action.yml
+++ b/download-center-upload/action.yml
@@ -37,7 +37,7 @@ runs:
   - name: Setup python
     uses: actions/setup-python@v5
     with:
-      python-version: '3.11'
+      python-version: '3.11' # gsutil requires python 3.5-3.11 (https://cloud.google.com/storage/docs/gsutil_install#specifications)
 
   - name: Resolve variables
     shell: bash
--- a/download-center-upload/action.yml
+++ b/download-center-upload/action.yml
@@ -34,6 +34,11 @@ runs:
   using: "composite"
 
   steps:
+  - name: Setup python
+    uses: actions/setup-python@v5
+    with:
+      python-version: '3.11'
+
   - name: Resolve variables
     shell: bash
     id: variables


barmac:
Fix verified via https://github.com/camunda/camunda-modeler/actions/runs/11379519741/job/31657108665

Langleu:
@barmac, neither Chaima nor I are in the team anymore.

barmac:
Sorry, I just noticed you in the suggestions and assumed GH is right.

kellervater (approved):
LGTM :shipit:

maxdanilov (changes requested):
I have concerns about not leaving any comment in the code on why 3.11 is the version to be pinned. This will create a maintenance confusion in the future.

barmac:
> I have concerns about not leaving any comment in the code on why 3.11 is the version to be pinned. This will create a maintenance confusion in the future.

I'll add it then.

barmac:
Comment added via https://github.com/camunda/infra-global-github-actions/pull/279/commits/1b3b940567a8196fbdd482abe83c6daaf6bbf040

BTW Google recommends to use another tool for storage: 
> Important: gsutil is not the recommended CLI for Cloud Storage. [Use gcloud storage](https://cloud.google.com/storage/docs/discover-object-storage-gcloud) commands in the Google Cloud CLI instead.

(source: https://cloud.google.com/storage/docs/gsutil)

kellervater:
@barmac 
I'll give `gcloud` storage a shot here and see if it's working with the current python version.

We assume that the `ubuntu-latest` runners, which just switched to `24.04` also have a different python version now. This may have caused the incompatibility.

kellervater:
@barmac created another PR which is using `gcloud storage` instead: 
https://github.com/camunda/infra-global-github-actions/pull/281

barmac:
Then I am closing this one. Thanks!

PR #280, opened 2024-10-17
feat(ci): add automerge action

kellervater:
With https://github.com/camunda/team-infrastructure/issues/614 we want to introduce a `Autorelease` feature.
One part is an automerge functionality, which is capable of automatically merging PRs based on label filters.

For this we introduced a new action group `pull-request` with 2 new actions:
* `release-please` which abstracts the call of the release-please action (create & update release PRs as well as creating new Github releases)
* `automerge`: Merges PRs with a given label (defaults to `autorelease: pending`)


pull-request/release-please/action.yml
```diff
@@ -0,0 +1,45 @@
+---
+name: Manage Release
+description: |
+  This action manages auto-release handling for the repository.
+  Based on the configuration in the release-please-config.json file in conjunction with
+  all conventional commits since the last release, it will create a new release PR.
+  If the PR itself gets merged, the action will create a new release and tag it.
+  The config-filename is required until https://github.com/googleapis/release-please-action/issues/1017 is resolved.
```
maxdanilov, line 8: I don't understand the logic here - if there is no other ways to get the file instead of reading it from the current repo, why we need to require the filename? opting for a default config filename in all the repositories would make less parameters to pass.

pull-request/automerge/README.md
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      # you can skip the create-github-app-token step, if using a PAT instead of a Github App
+      - name: Create Github Token
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: 1234567 # make sure the app can modify PRs, contents and actions
+          private-key: ${{ secrets.APP_PRIVATE_KEY }} # needs to be set on repo or orga level
```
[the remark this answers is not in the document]
kellervater: oh yeah. that slipped through. Good catch

pull-request/release-please/action.yml
```diff
@@ -0,0 +1,44 @@
...
+  manifest-filename:
+    description: |
+      filename of a custom release-please manifest file.
+    required: false
+    default: '.github/config/.release-please-manifest.json'
+
+runs:
+  using: composite
+  steps:
+    - name: config-filename defaulting
```
maxdanilov, line 31: same here (relative zero indent)

pull-request/automerge/README.md
```diff
@@ -0,0 +1,31 @@
...
+permissions:            # Additional permissions for standard GITHUB_TOKEN
+  contents: write       # so we can modify the changelog
+  pull-requests: write  # so release-please can create/update PRs
+  actions: write        # so subsequent actions can run
+
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create Release
```
maxdanilov, line 26: same here (relative zero indent)

pull-request/automerge/action.yml
```diff
@@ -0,0 +1,41 @@
+---
+name: Automerge Pull Requests
+description: |
+  Automatically merge pull requests.
+  The action fetches all open pull requests with a specific label (default: `autorelease: pending`) and merges them.
```
maxdanilov, line 5: why not use the existing approach in https://github.com/camunda/infra-core/blob/4fa292ff8293c95ad50aab7838b7722ebd37f5e5/.github/workflows/automerge-master-pr.yml#L2?

maxdanilov (changes requested):

pull-request/automerge/action.yml
kellervater: hmm... I really tried several things now, but I can't get it working with the proposed solution:
https://github.com/kellervater/c8sm/actions/runs/11463419684/job/31897218278

it just doesn't recognize the [release-please PR](https://github.com/kellervater/c8sm/pull/38) in my test repository. Already put some time into that. Not sure if I'm missing something super obvious here.
but tbh. I feel like I'm currently wasting a lot of time here.

kellervater: tried different labels because I thought the space of `autorelease: pending` label is the problem.
Tried a direct call instead of the wrapper because I thought there may be a contextual problem. nothing helped so far.

maxdanilov: @kellervater can you invoke the workflow in a debug mode?

kellervater: yes. but it just doesn't tell me WHY it's not picking up on the PR. I'm currently trying different parameter combinations.

here a debug log: https://github.com/kellervater/c8sm/actions/runs/11463714585/job/31898205486

kellervater: Also an almost 1:1 copy if infra-core's automerge workflow doesn't bring the expected result:
https://github.com/kellervater/c8sm/actions/runs/11463898128/job/31898836829

I suspect, that it doesn't work, when not triggered by a pull request event.

kellervater: it may also be, because the label gets introduced by another bot. There's an issue hinting at that (but with some different use-case though):
https://github.com/pascalgn/automerge-action/issues/131

maxdanilov: I think it's because the action behaves differently based on what trigger happened: https://github.com/pascalgn/automerge-action/blob/f981436038b98090c5ad6b89718eeee5baa9847f/lib/api.js#L90

The workflow dispatch one calls this handler, with the Octokat PR filter looking suspicious (where's the repo name?):  https://github.com/pascalgn/automerge-action/blob/f981436038b98090c5ad6b89718eeee5baa9847f/lib/api.js#L249

kellervater: I'm currently unable to fix that problem with the third party action in a reasonable time (maybe too late already).
I could try to debug deeper tomorrow, but tbh: There was a quite simple and working solution in the first place, which should do just fine.
@maxdanilov do you think it's worth investigating further? Then I'll continue to look for a solution on the third-party automerge action tomorrow.

kellervater: > The workflow dispatch one calls this handler, with the Octokat PR filter looking suspicious (where's the repo name?): https://github.com/pascalgn/automerge-action/blob/f981436038b98090c5ad6b89718eeee5baa9847f/lib/api.js#L249

hmm... good hint. will try to debug further

maxdanilov: Looking into the REST API [doc](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests) of GitHub, I'm 100% sure that line is responsible one. The `head` value it passes looks like `kellervater:main` and the filter returns nothing. To capture the [PR](https://github.com/kellervater/c8sm/pull/38), the `head` value should've been `kellervater:release-please--branches--main`. Looks like a bug in the code (and I don't see why the `head` filter is needed in this case at all).

maxdanilov: Before debugging, try just whether it works properly with the cronjob event - if that is the main one, this may be sufficient already.

kellervater: > Try with the cronjob trigger, looking at the code it should work.

did that: https://github.com/kellervater/c8sm/actions/runs/11464405607/job/31900517435
no change

will continue tomorrow with an in-depth debugging session then :-)

kellervater: or not. it seems like it recognized the last one as workflow_dispatch again. strange.

maxdanilov: worked https://github.com/kellervater/c8sm/actions/runs/11464676291 🎉  just needs `MERGE_REQUIRED_APPROVALS` unspecified or 0.

kellervater: you were right. This time  it at least found the PR: https://github.com/kellervater/c8sm/actions/runs/11464676291/job/31901404272

Maybe worth opening an issue at the vendor, so they can fix the workflow_dispatch behavior (or do it ourselves and hope for a merge). 

Will try to continue with this approach now.

kellervater: the third(rty) time is the charm ;-)
https://github.com/kellervater/c8sm/actions/runs/11464837600/job/31901942700

thx @maxdanilov for your advise. Will use this approach now :-)

teams/infra/pull-request/automerge/action.yml
```diff
@@ -0,0 +1,27 @@
...
+    default: 'automerge'
+
+runs:
+  using: composite
+  steps:
+  - name: automerge
+    uses: pascalgn/automerge-action@v0.16.4
+    env:
+      GITHUB_TOKEN: "${{ inputs.github-token }}"
+      MERGE_METHOD: "squash"
```
maxdanilov, line 26: (non-blocking)
what was the reasoning for going with `squash` vs `merge` and preserving the git history?

teams/infra/pull-request/automerge/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Create Release
+      uses: camunda/infra-global-github-actions/pull-request/automerge@main
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        label: <yourCustomLabel> # optional, defaults to 'autorelease: pending'
```
maxdanilov, line 35: doesn't it default to `automerge`?

maxdanilov (changes requested):

teams/infra/pull-request/automerge/action.yml
kellervater: It was a coincidence that I discovered it:
If a repositories branch protection is set to "require signed commits" then "merge" doesn't work anymore since a merge commit seems to remove the signature, whereas a `squash` seems to work just fine. At least on my testing repository. We can also go with "merge" or "rebase", but I also like the default squash message which refers to the release PR.

maxdanilov (approved):

PR #285, opened 2024-10-23
chore(rerun-failed-run): check for errors in failure annotations

clementnero:
Related to https://github.com/camunda/team-infrastructure/issues/707.

Specified error messages are checked in annotations of `FAILURE` level. A new `rerun-failed-run/check-annotations-and-logs` composite action has been introduced to better abstract the logic.

When the `rerun-failed-run` composite action triggers a new run of `retry-worfklow-run.yml`, it looks up the run id and related url, in order to create a new `warning` annotation for better visibility and easier debugging.

DEBUG logs have been removed, and logs improved.


.github/workflows/retry-worfklow-run.yml
```diff
@@ -0,0 +1,138 @@
...
+        required: false
+
+jobs:
+  retry:
+    name: Retry workflow run
+    runs-on: ubuntu-latest
+    steps:
+      - name: Distinct Id to retrive this worfklow run ${{ inputs.distinct_id }}
+        run: |
+          echo ${{ inputs.distinct_id }}
```
kellervater, line 47: non-blocking: 
if the `distinct_id` is unset, this step has an empty output which might be a bit confusing when looking at job logs.

clementnero: I will add a default value then.

clementnero: Done.

rerun-failed-run/action.yml
```diff
@@ -52,17 +52,53 @@ runs:
...
+        messages_json=$(
+          echo "${ERROR_MESSAGES}" | \
+            jq -c -n -R '[inputs | select(length > 0)]' | \
+            jq -R .
+        )
+        echo json="${messages_json}" | tee -a $GITHUB_OUTPUT
+      shell: bash
 
-        gh workflow run rerun-failed-run.yml -R camunda/infra-global-github-actions --ref=main -F run_id=${{ inputs.run-id }} -F repository=${{ inputs.repository }} -F error_messages="$result" -F notify_back_on_error=${{ inputs.notify-back-on-error }} -F notify_back_git_ref=${{ github.ref_name }}
+    - name: Trigger ${{ github.repository }}/retry-worfklow-run.yml
```
kellervater, line 78: I am not sure if I understand correctly, but here it looks like the original repository must implement `retry-workflow-run`. 

Are there cases where this is not `infra-global-github-actions`? If so, we should describe that in the accompanying readme I guess.

rerun-failed-run/check-annotations-and-logs/find-errors.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Exit immediately if a command exits with a non-zero status
+set -e
```
kellervater, line 4: for a fail-fast approach may also a `-u` flag would make sense.

kellervater (approved):
LGTM in general. Looks very clean. 
Left just one question about eventually necessary documentation.

rerun-failed-run/action.yml
clementnero: Good catch! It should not be dynamic indeed, and always call camunda/infra-global-github-actions/retry-workflow-run.yml.
Since my test was on the same repo it worked and I did not spot that issue.

clementnero: Fixed.

rerun-failed-run/check-annotations-and-logs/find-errors.sh
clementnero: It makes sense :+1

clementnero: I will add all the options ^^, thx!

clementnero: Fixed.

PR #311, opened 2024-12-11
ci: fix infinite loop in actionlint

Kerruba:
Related to: camunda/team-infrastructure#620


kellervater (commented):
Did you test this somewhere?
Just want to avoid that we need to reiterate again.

kellervater (approved):
as discussed with the author: tested locally on dev workstation.

Kerruba:
Here a test on a real case scenario: https://github.com/camunda/camunda/actions/runs/12279941772/job/34265221798
Previously this wasn't working due to the infinite loop, with this change the infinite loop is gone

PR #316, opened 2024-12-16
chore(feat): enable using google workload identity

kalmarz:
Skip authenticating with a Docker registry using username/password in case of the caller workflow uses Google Workload Identity.

Related:
camunda/team-infrastructure/issues/629


kellervater (approved):
LGTM, thank you.

PR #318, opened 2024-12-19
Camunda 24303 fix action path

Kerruba:
This PR is for implementing the cache disable mechanism but in a way that actually works when referencing the `is-cache-enabled` composite action

### DIFF
--- a/is-cache-enabled/action.yml
+++ b/is-cache-enabled/action.yml
@@ -1,0 +1,43 @@
+---
+name: Is cache enabled
+
+description: |
+  Checks if cache is enabled given three conditions
+  - is a pull request related event
+  - the pull request does not contains a 'ci:no-cache' label
+  - the branch is neither 'main' nor 'stable/*'
+  This information can be used to skip cache restoration
+  in the setup-maven-cache and setup-yarn-cache to test
+  workflows from scratch
+
+inputs: {}
+
+outputs:
+  is-cache-disabled:
+    description: |
+      Output whether the workflows has cache enabled based on three conditions:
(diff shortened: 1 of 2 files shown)


setup-yarn-cache/action.yml
```diff
@@ -38,6 +38,17 @@ runs:
         echo "result=false" >> $GITHUB_OUTPUT
       fi
 
+  - name: Symlink current action repo
+    env:
+      action_path: ${{ github.action_path }}
+    run: ln -fs ${{ env.action_path }}/.. .github/.linked-infra-global-github-actions
+    shell: bash
```
cmur2, line 45: This workaround is necessary as per https://github.com/orgs/community/discussions/41927

is-cache-enabled/action.yml
```diff
@@ -0,0 +1,43 @@
...
+runs:
+  using: composite
+  steps:
+
+  - id: detect-cache-status
+    shell: bash
+    run: |
+      IS_PR_EVENT=${{ github.event_name == 'pull_request' }}
+      HAS_NO_CACHE_LABEL=${{ contains(github.event.pull_request.labels.*.name, 'ci:no-cache') }}
+      IS_FEATURE_BRANCH=${{ !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main') }}
```
cmur2, line 33: Introduce and respect the `cache_create_branch_regex` input that `setup-yarn-cache` action has - to allow universal use also outside of `camunda/camunda` repo.

cmur2 (changes requested):

PR #336, opened 2025-01-14
chore(submit-test-status): disable echoing potentially long test record input

cmur2:
Small change I noticed while working on https://github.com/camunda/camunda/pull/26715 that showing the test record input can potentially result in many 100 lines of output. I'd propose to not echo this.


kalmarz:
Had used it for "printf debugging" and forgot to remove.

kalmarz (approved):

PR #343, opened 2025-01-22
chore: setup automerge for actionlint patch updates

Kerruba:
In camunda's renovate shared configuration, actionlint patch updates are automerged by default. 
Adding the same rule here to simplify maintenance

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -73,6 +73,6 @@
       ],
       addLabels: ["automerge"],
       automerge: true,
-    } 
+    }
   ]
 }
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -56,6 +56,14 @@
       addLabels: ["automerge"],
       automerge: true
     },
+    {
+      matchUpdateTypes: ["patch"],
+      matchPackageNames: [
+        "/actionlint/"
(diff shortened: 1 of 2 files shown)


kellervater (changes requested):
First of all: Good catch!  Really perceptive that you noticed that this dependency got automerged somewhere else.

For the future: Could you please link the central config and a reference PR which got automerged using said config?

Something like:
```
Central Config for comparison: https://github.com/camunda/infra-renovate-config/blob/main/default.json5
Example PR which validates this config: https://github.com/camunda/infra-core/pull/9008
```

leaving a small remark. otherwise lgtm.

Kerruba:
Thanks @kellervater for your suggestions,
Here the links for future references, so that even this one has the correct format

Centralized Configuration: https://github.com/camunda/infra-renovate-config/blob/ecf86195c57e5a655ee2d97c4d1bb65421658204/default.json5#L219

Other PRs automerged:
- https://github.com/camunda/infra-core/pull/9008
- https://github.com/camunda/infra-preview-environments-ingress/pull/112

I'm now going to move the section as suggested

kellervater (approved):
LGTM, thanks for this improvement

PR #351, opened 2025-02-03
ci(renovate): increase automerge rate for argocd

kellervater:
ArgoCD patch releases should be automerged (see [this PR](https://github.com/camunda/infra-core/pull/9070) as reference).
Didn't happen though (see [this PR](https://github.com/camunda/infra-global-github-actions/pull/348)).

Adapted the config according to our [central one](https://github.com/camunda/infra-renovate-config/blob/main/default.json5#L232).

After merge I will retry the yet unmerged PR to validate.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -56,10 +56,12 @@
       addLabels: ["automerge"],
       automerge: true
     },
+    // Automerge Section
     {
       matchUpdateTypes: ["patch"],
       matchPackageNames: [
-        "/actionlint/"
+        "/actionlint/",
+        "argoproj/argo-cd",
       ],
       addLabels: ["automerge"],
       automerge: true,


kalmarz (commented):
I'm wondering if getting rid of this whole automerge section makes sense?
As far as I see there's nothing specific to it.
It'd just inherit everything from the main renovate config.
Also see our discussion in: https://camunda.slack.com/archives/C087RKSFCR1/p1737970089171309?thread_ts=1737968805.914829&cid=C087RKSFCR1

kellervater:
Problem is, that the central renovate config is private/internal. That's why this public repo cannot inherit from the central one. I already brought that up [here](https://camunda.slack.com/archives/CHY2S7KDJ/p1734340888402549). 
Since it's not very often the case that we have to update the renovate config, I think it's fine for now.

kalmarz (approved):

PR #359, opened 2025-02-28
Install envsubst

kalmarz:
Install `envsubst` binary for `preview-env/create` action. When the action is executed on a self-hosted runner then the deployment artifact create step would fail.

- Related
  - https://github.com/camunda/team-infrastructure/issues/521
  - https://camunda.slack.com/archives/C5AHF1D8T/p1740729769640149

- Tested on `web-modeler`
  - https://github.com/camunda/web-modeler/actions/runs/13586248134/job/37981769877?pr=12687#step:7:1391

### DIFF
--- a/preview-env/create/action.yml
+++ b/preview-env/create/action.yml
@@ -187,6 +187,13 @@ runs:
       env_url: ${{ inputs.app_url }}
       ref: ${{ inputs.revision }}
       env: ${{ inputs.app_name }}
+    # envsubst is not available on the self-hosted runners
+    # therefore the deployment status artifact step would fail
+  - name: Install envsubst
+    if: always()
+    shell: bash
+    run: |
+      sudo apt-get update && sudo apt-get install -y gettext-base
   - name: Create deployment status artifact
     if: always()
     shell: bash


clementnero (approved):

preview-env/create/action.yml
```diff
@@ -187,6 +187,13 @@ runs:
       env_url: ${{ inputs.app_url }}
       ref: ${{ inputs.revision }}
       env: ${{ inputs.app_name }}
+    # envsubst is not available on the self-hosted runners
+    # therefore the deployment status artifact step would fail
+  - name: Install envsubst
+    if: always()
+    shell: bash
+    run: |
+      sudo apt-get update && sudo apt-get install -y gettext-base
```
maxdanilov, line 196: This workflow is not guaranteed to run on self-hosted runners, can be both self-hosted and SaaS ones.
So sometimes it can be envsubst is installed and does not need to be installed again.
So I'd say there should be an idempotency approach:
- check `envsubst` is installed
- only install if it's not there

clementnero: The test mentioned above on web-modeler, https://github.com/camunda/web-modeler/actions/runs/13586248134/job/37981769877?pr=12687#step:7:1391, run on GitHub-Managed runners, and in this case it just try to update the package, so no immediate issue.

```bash
gettext-base is already the newest version (0.21-14ubuntu2).
gettext-base set to manually installed.
```

clementnero: Overall yeah, good catch!

PR #365, opened 2025-03-10
chore(deps): enable renovate automerge for teleport-ent-distroless

kalmarz:
Enable Renovate automerge for `gravitational/teleport-ent-distroless` patch releases.

Related: https://github.com/camunda/infra-global-github-actions/pull/362

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -62,6 +62,7 @@
       matchPackageNames: [
         "/actionlint/",
         "argoproj/argo-cd",
+        "gravitational/teleport-ent-distroless",
       ],
       addLabels: ["automerge"],
       automerge: true,


kellervater (commented):
Have you tested this somewhere to validate is picked up for automerge?

kalmarz:
> Have you tested this somewhere to validate is picked up for automerge?

I have not. Just followed the argocd pattern that found one line above. What's the best way to test Renovate updates?

kellervater:
I normally replicate the respective section which would be updated by renovate in `infra-github-demo` (see section renovate), then adapt the renovate config there and see if the respective PR would receive "Automerge: Enabled".

Otherwise you risk YOLOing the change which might not even have any effect. We had that a lot last year.

kalmarz:
This action has been used only in a single workflow so far. From the effort - benefit perspective I'd just YOLO it for now. It doesn't do any bad anyway, and if comes back at the next patch version, we can get back to the topic.

kellervater (approved):
fair enough ;-)

PR #373, opened 2025-03-26
chore: add fossa-related composite actions

clementnero:
Related to https://github.com/camunda/team-infrastructure/issues/751.

Add new composite actions for FOSSA:
- `fossa/setup` to install the CLI
- `fossa/analyze` to run the analysis

The idea is to encapsulate interactions with the FOSSA CLI used in workflows across various product repositories. This approach simplifies maintenance (e.g., upgrading the CLI version) and abstracts unnecessary implementation details.

### DIFF
--- a/fossa/analyze/action.yml
+++ b/fossa/analyze/action.yml
@@ -1,0 +1,40 @@
+name: fossa analyze
+
+description: Run FOSSA CLI analyze command
+
+inputs:
+  api-key:
+    description: The API key to access fossa.com
+    required: true
+  branch:
+    description: Name of the branch to be scanned
+    required: true
+  configuration-file:
+    description: Path to the FOSSA configuration file
+    required: false
+    default: .fossa.yml
+  path:
+    description: Path to the directory to be scanned
+    default: .
+    required: false
+  revision-id:
+    description: ID of the scanned revision
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Run FOSSA analyze
+      env:
+        BRANCH: ${{ inputs.branch }}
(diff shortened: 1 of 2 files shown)


fossa/analyze/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    required: true
+  configuration-file:
+    description: Path to the FOSSA configuration file
+    required: false
+    default: .fossa.yml
+  path:
+    description: Path to the directory to be scanned
+    default: .
+    required: false
+  revision-id:
```
kellervater, line 20: i wonder if we could just default this to the current branch HEAD ?

kellervater (approved):
LGTM :rocket: 
short, sweet, lean.
I like it. 
Left 2 minor questions, but feel free to merge without taking them into consideration.

fossa/analyze/action.yml
clementnero: Depending of the type of event, HEAD may be retrieved from different attributes of the GitHub context (push vs pull-request). I prefer to keep it explicit for now, at least until the implementation of PR checks.

fossa/setup/action.yml
```diff
@@ -0,0 +1,29 @@
+name: fossa setup
+
+description: Install FOSSA CLI
+
+inputs:
+  version:
```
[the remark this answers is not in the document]
clementnero: It's not really necessary but I think it makes sense to allow user to set a different version (kind of a generic approach). Also it allows to clearly surface that the default version is tracked by renovate (it's only recommended to use the latest version, no compatibility matrix for the CLI is provided).

PR #399, opened 2025-05-07
feat: add new fossa-related composite actions

clementnero:
Related to https://github.com/camunda/team-infrastructure/issues/752.

Add new fossa-related composite actions to check PRs for newly introduced issues. It currently only handle license issues, but can be extended to security issues later.

The `fossa/pr-check` action relies on the `fossa test` command and includes a License Policy guard step, which fails if the PR introduces a new license issue and add an annotation to the job for visibility.

### DIFF
--- a/fossa/pr-check/action.yml
+++ b/fossa/pr-check/action.yml
@@ -1,0 +1,79 @@
+name: pr check
+
+description: |
+  Check for new license issues introduced by a pull request.
+
+inputs:
+  api-key:
+    description: The API key to access fossa.com
+    required: true
+  base-ref:
+    description: |
+      The ref (name) of the base branch in the context of a pull request.
+      Only for display purposes, not required.
+    required: false
+  base-revision:
+    description: |
+      The revision (commit SHA) of the base branch in the context of a pull request.
+    required: true
+  configuration-file:
+    description: Path to the FOSSA configuration file
+    required: false
+    default: .fossa.yml
+  path:
+    description: Path to the directory for scanning
+    default: .
+    required: false
+  project:
+    description: |
+      The project name.
+      Only for display purposes, not required.
+    default: main
+    required: false
+  revision:
+    description: |
+      The revision (commit sha) of the HEAD branch in the context of a pull request.
+    required: true
+
+runs:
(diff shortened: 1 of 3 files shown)


fossa/pr-check/action.yml
```diff
@@ -0,0 +1,81 @@
+name: pr check
+
+description: |
+  Check for new issues introduced by a pull request.
+
+  Note: Currently limited to license-related issues.
```
kellervater, line 6:
```suggestion
  Check for new license issues introduced by a pull request.
```
Would make it more concise.

fossa/pr-check/action.yml
```diff
@@ -0,0 +1,81 @@
...
+      Only for display purposes, not required.
+    default: main
+    required: false
+  revision:
+    description: |
+      The revision (commit sha) of the HEAD branch in the context of a pull request.
+    required: true
+
+runs:
+  using: composite
```
kellervater, line 41: I was wondering: Did you also consider using the 1st party [Github Action](https://github.com/fossas/fossa-action) provided by Fossa. They also seem to provide a [test-diff-version](https://github.com/fossas/fossa-action?tab=readme-ov-file#test-diff-revision) function which might be sufficient.

kellervater (approved):
Nice work :-D 
I like the annotation approach.
Left a clarification question about the reasoning to go with a custom implementation.
If my concerns are non-applicable, please feel free to merge this PR as it is a well-crafted addition to our global actions.

fossa/pr-check/action.yml
clementnero: Good question. Yes, I considered it, but it doesn't support all the parameters we use with the FOSSA CLI, and that was a blocker. This approach also gives us full control over the interface, which is especially useful when integrating multiple repositories as we do.

Creating these composite actions, and so an interface, doesn’t prevent us from using 1st party gha (behind the interface) in the future if it ends up being a better fit.

fossa/pr-check/action.yml
clementnero: Indeed.

PR #410, opened 2025-05-22
feat(gha): enable retry action to be used without error-message

leiicamundi:
This PR enables us to retry a workflow without checking the error-message.

This can be useful when we want to ensure that the workflow is always reaching a termination point (maximum number of retries) so it can integrates cleanup steps (if  fromJSON(github.run_attempt) >= 3 => triggers final cleanup).

This PR also fixes: a bad interpolation of the environment variable due to usage of `-`, fixed by

```
notify_back_error_message="${NOTIFY_BACK_ERROR_MESSAGE}" \
              --ref="${notify-back-git-ref}"
              --ref="${{ inputs.notify-back-git-ref }}"
```


clementnero (approved):
LGTM, thank you 🚀

Langleu (commented):
Not the owner of the action, but makes sense to me 👍.
Thanks for fixing the interpolation problem!

PR #417, opened 2025-06-02
feat(fossa): add fallback to recent commits on base ref if first commit has no FOSSA scan

clementnero:
Related to https://github.com/camunda/team-infrastructure/issues/752 and https://camunda.slack.com/archives/C5AHF1D8T/p1748855183016529.

The` fossa/pr-check `GitHub Action compares FOSSA scan results between the head and base branches of a pull request to detect newly introduced license issues. Sometimes, the first commit on the base ref may not have a FOSSA scan, for example, due to workflows not being triggered (e.g. if commit pushed from a workflow) or interrupted, outages, etc.

To handle this, the action now includes a fallback using a new output: `base-revision-most-recent-with-scanning-results`. If the first base commit lacks scan results, the action checks up to the 5 most recent commits on the base branch for the latest one with scan results. For merge commits, parent 1 (the branch that was merged into) is used to preserve context. If no commits are found, `base-revision-most-recent-with-scanning-results` returns empty.

In rare cases, this might misattribute existing (not resolved) license issues from the base branch to the PR. However, this risk is very low. This implementation let the user decide whether to use this new output (fallback) or not (I will update all repositories).

This also lays the groundwork for making the 5-commit threshold configurable in the future, if needed.

Please find below the different tests (PRs and workflow runs) on a demo repository:
- TEST [01](https://github.com/clementnero/fossa-demo-case-01/actions/runs/15399182930/job/43327383868?pr=8): The base ref has a FOSSA scan result, check-licenses.yml runs as expected
- TEST [02](https://github.com/clementnero/fossa-demo-case-01/actions/runs/15399461759/job/43328335694?pr=9): The base ref has no scan result, but a recent commit on the base branch does, `check-licenses.yml `falls back and runs as expected
- TEST [03](https://github.com/clementnero/fossa-demo-case-01/actions/runs/15399604762/job/43328968516?pr=10): Neither the base ref nor recent commits have scan results, `check-licenses.yml` fails as expected.
- TEST [04](https://github.com/clementnero/fossa-demo-case-01/pull/11/checks): For a merge commit with no workflow run (i.e., no license scan), the action successfully falls back to recent scanned commits (similar scenario as https://camunda.slack.com/archives/C5AHF1D8T/p1748855183016529)

### DIFF
--- a/fossa/info/action.yml
+++ b/fossa/info/action.yml
@@ -2,6 +2,12 @@ name: fossa info
 
 description: Provides context info required for other FOSSA composite actions.
 
+inputs:
+  github-token:
+    description: |
+      The GitHub token to use for API calls.
+    default: ${{ github.token }}
+
 outputs:
   is-pull-request:
     description: |
@@ -18,6 +24,12 @@ outputs:
       Empty if not a pull request.
       Generally used to find (via diff) new license issues introduced by a PR.
     value: ${{ steps.info.outputs.base-revision }}
+  base-revision-most-recent-with-scanning-results:
+    description: |
+      The latest base branch revision (commit SHA) with FOSSA scanning results.
+      Useful when the base revision has no scanning results, such as after a manual merge commit that did not trigger a workflow (and so a scan).
+      Empty if none found.
+    value: ${{ steps.info-pr.outputs.base-ref-with-scanning-results }}
   head-ref:
     description: |
       The HEAD ref (name) of the branch to be analyzed by FOSSA, determined based on the context of the event.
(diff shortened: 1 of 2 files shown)


fossa/info/action.yml
```diff
@@ -2,6 +2,12 @@ name: fossa info
 
 description: Provides context info required for other FOSSA composite actions.
 
+inputs:
+  github-token:
+    description: |
+      The GitHub token to use for API calls.
+    required: true
+
```
kellervater, line 10: Am I right when saying, that this is a breaking change, when introduced?
I think we should therefore coordinate the merge of this and the subsequent PRs to avoid confusion amongst teams.

fossa/info/action.yml
```diff
@@ -67,3 +79,38 @@ runs:
...
+    if: ${{ steps.info.outputs.is-pull-request == 'true' }}
+    id: info-pr
+    env:
+      BASE_REF: ${{ steps.info.outputs.base-ref }}
+      BASE_REVISION: ${{ steps.info.outputs.base-revision }}
+      GH_TOKEN: ${{ inputs.github-token }}
+    run: |
+      # Check recent commits of the base branch for existing scanning results
+      # history depth to search for scanning results
+      depth=5
```
kellervater, line 92: maybe a bit overkill but: Could it make sense to make this configurable as another input which defaults to 5?

fossa/info/action.yml
```diff
@@ -67,3 +79,38 @@ runs:
...
+      # Ensure history is available for the base branch up to the specified depth
+      git fetch origin "${BASE_REF}" --depth="$depth"
+      # Check if the base branch or its recent parents have scanning results
+      base_ref_with_scanning_results=""
+      for i in $(seq 0 $((depth-1))); do
+        commit=$(git rev-parse "${BASE_REVISION}~$i" 2>/dev/null)
+        # Check if a workflow run of "Check Licenses" succeeded for the commit
+        if gh run list --commit "$commit" --json workflowName,conclusion \
+          --jq '[.[] | select(.workflowName == "Check Licenses" and .conclusion == "success")] | any' | grep -q true; then
```
kellervater, line 101: can we guarantee that the workflow name for the fossa check is always equal to "Check Licenses"?
If we want to rely on that we should definitely document that in the calling workflows with some kind of warning, so teams are aware of that.

Maybe also overkill and wildly inefficient, but in a combination with `gh run view <run-id> --log` and some grep magic we could get info if the action has actually been called?

kellervater: Or maybe even easier: The context should have the workflow name available as `${{ github.workflow }}`

fossa/info/action.yml
```diff
@@ -67,3 +79,38 @@ runs:
...
+          if [ "$i" -gt 0 ]; then
+             echo "Scanning result found for commit=$commit, which is $i commit(s) behind the base branch/ref."
```
kellervater, line 103: do you think that's worth annotating in the github job for visibility?

fossa/pr-check/action.yml
```diff
@@ -50,6 +50,11 @@ runs:
       # Run fossa test
       results=$(fossa test "${DIRECTORY_PATH}" --config "${CONFIGURATION_FILE}" --diff "${BASE_REVISION}" --revision "${REVISION}" --format json || true)
 
+      # Exit if results is empty or not a valid JSON
```
kellervater, line 53:
```suggestion
      # Fail if results is empty or not a valid JSON
```

kellervater (approved):
LGTM :rocket: 

Great work! 
This is some really thorough testing work. I appreciate that very much!

I left some non-blocking nitpicks/suggestions/questions. You can dismiss them regardless.

fossa/info/action.yml
clementnero: Indeed, yes. I will create subsequent PRs.

fossa/info/action.yml
clementnero: Good catch! `${{ github.workflow }}` should do the trick

fossa/info/action.yml
clementnero: I don't think we want to advertise this outside of the logs for now, as this is no actionable for end-users (but noise).

fossa/info/action.yml
clementnero: No breaking regarding that part.

PR #422, opened 2025-06-11
feat: Introduce previous-version GHA (32376)

oleksandr-kriuchenko-lohika:
This GHA is supposed to be used for previous release version idebtification based on the formalized rules (covered with tests) and the existing git tags. Can be used in any repository, but primarly targets monorepo and connectors (from where the script originates).

Usage test: https://github.com/camunda/infra-global-github-actions/actions/runs/15587631297/job/43897949314

### DIFF
--- a/previous-version/previous_version.sh
+++ b/previous-version/previous_version.sh
@@ -16,10 +16,10 @@
 #   E.g. "NORMAL 8.7.0"
 
 PREV_TAG="not found"
-VERSION=$1
-VERBOSE=$2
+VERSION="$1"
+VERBOSE="$2"
 
-if [[ "${VERBOSE}" = "true" ]]; then
+if [[ ${VERBOSE} = "true" ]]; then
   set -x
 fi
 


previous-version/README.md
```diff
@@ -0,0 +1,40 @@
+# previous-version
+
+Given a release version identifies release type and a corresponding previous release version
```
maxdanilov, line 3: (minor suggestion)

```suggestion
Given a release version identifies release type and a corresponding previous release version in a current repository.
```

previous-version/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Usage
+
+This GHA can be used in any repository, but requires it to be fetched with full history,
+so that existing tags can be analyzed. Otherwise, results may be incorrect.
+
+### Inputs
+| Input name | Description                            |
+|------------|----------------------------------------|
+| version    | current release version (e.g. 8.8.0)   |
+| verbose    | provide verbose output (true \| false) |
```
maxdanilov, line 14: let's add that default value for `verbose` is `false` and that `version` is required.

previous-version/README.md
```diff
@@ -0,0 +1,40 @@
+# previous-version
+
+Given a release version identifies release type and a corresponding previous release version
+
+## Usage
+
+This GHA can be used in any repository, but requires it to be fetched with full history,
+so that existing tags can be analyzed. Otherwise, results may be incorrect.
+
```
maxdanilov, line 9: Let's also add similar table for outputs. Inputs + Outputs would fully describe the interface of this action.

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+PREV_TAG="not found"
+VERSION=$1
+VERBOSE=$2
+
+if [ "$VERBOSE" = "true" ]; then
+  set -x
+fi
+
+  # Function to determine the previous tag using 'NORMAL' logic
+  determine_normal_prev_tag() {
```
maxdanilov, line 27:
```suggestion
determine_normal_prev_tag() {
```

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+#   2 - Verbose mode (true|false)
+# Output:
+#   "$TYPE $PREV_TAG"
+#   E.g. "NORMAL 8.7.0"
+
+PREV_TAG="not found"
+VERSION=$1
+VERBOSE=$2
+
+if [ "$VERBOSE" = "true" ]; then
```
maxdanilov, line 22: in this repo, the convention is to use 2 spaces as tabulation, let's do the same in this file.

previous-version/previous_version-test.sh
```diff
@@ -0,0 +1,212 @@
+#!/bin/bash
+#
```
maxdanilov, line 2: same here for tabulation: let's use 2 spaces in this file.

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+# Given a current version identifies a previous version based on existing git tags
+# Params:
+#   1 - Current version. E.g. 8.8.0
+#   2 - Verbose mode (true|false)
+# Output:
+#   "$TYPE $PREV_TAG"
+#   E.g. "NORMAL 8.7.0"
+
+PREV_TAG="not found"
+VERSION=$1
```
maxdanilov, line 19:
```suggestion
VERSION="${1}"
```

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+VERBOSE=$2
```
maxdanilov, line 20:
```suggestion
VERBOSE="${2}"
```

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+  determine_normal_prev_tag() {
+      local candidate=""
+      local candidateWithPatch0=""
+      # Loop through all previous tags to find the appropriate previous tag
+      for tag in $prev_tags; do
+          if [[ $tag != "$current_tag" ]] && [[ $tag =~ $NORMAL_PATTERN ]]; then
+              IFS='.' read -r tag_major tag_minor tag_patch <<< "$tag"
+              IFS='.' read -r curr_major curr_minor curr_patch <<< "$current_tag"
+              curr_patch=${curr_patch%%-*}
+               if { ((tag_major == curr_major)) && ((tag_minor == curr_minor)) && ((tag_patch == curr_patch)); }; then
```
maxdanilov, line 36: I'm confused with this syntax and not sure it would work.
What about:
```suggestion
               if (( tag_major == curr_major && tag_minor == curr_minor && tag_patch == curr_patch )); then
```

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+  if [[ $type == "RC" ]]; then
+    # Extract the base version and RC version from the current tag
+    local base_version="${current_tag%-rc*}"
+    local -r current_rc_version=$(echo "$current_tag" | grep -oE "rc[0-9]+" | sed 's/rc//')
+
+    # Get the previous RC tag within the same minor version and with a lower RC number
+    prev_tag=$(get_previous_rc_tag "$base_version" "$current_rc_version" "$prev_tags")
+  fi
+
+  if [[ $type == "ALPHA" ]] || { [[ -z $prev_tag ]] && [[ $current_tag =~ -alpha[0-9]+(\.[0-9]+)?-rc[0-9]+$ ]]; }; then
```
maxdanilov, line 94: Also not sure about the syntax, what about:
```suggestion
  if [[ $type == "ALPHA" ]] || ( [[ -z $prev_tag ]] && [[ $current_tag =~ -alpha[0-9]+(\.[0-9]+)?-rc[0-9]+$ ]] ); then
```

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+        tag_rc_version=$(echo "$tag" | grep -oE "rc[0-9]+" | sed 's/rc//')
+        if (( tag_rc_version < current_rc_version )) && (( tag_rc_version > candidate_rc_version )); then
+          candidate_tag="$tag"
+          candidate_rc_version=$tag_rc_version
+        fi
+      fi
+    done
+    echo "$candidate_tag"
+}
+# Function to get the previous alpha tag
```
maxdanilov, line 130: two new lines are missing

previous-version/previous_version.sh
maxdanilov, line 22: A recommended syntax is:
```suggestion
if [[ "${VERBOSE}" = "true" ]]; then
```

previous-version/previous_version.sh
```diff
@@ -0,0 +1,189 @@
...
+    local tag_alpha_version
+    local tag_alpha_lt_current_alpha
+    local tag_alpha_gt_candidate_alpha
+
+    for tag in $prev_tags; do
+      if [[ $tag == $base_version-alpha* ]] && [[ $tag =~ $ALPHA_PATTERN ]]; then
+        tag_alpha_version=$(echo "$tag" | grep -oE "alpha[0-9]+(\.[0-9]+)?" | sed 's/alpha//')
+        tag_alpha_lt_current_alpha=$(echo "$tag_alpha_version < $current_alpha_version" | bc)
+        tag_alpha_gt_candidate_alpha=$(echo "$tag_alpha_version > $candidate_alpha_version" | bc)
+        if [ "$tag_alpha_lt_current_alpha" -eq 1 ] && [ "$tag_alpha_gt_candidate_alpha" -eq 1 ]; then
```
maxdanilov, line 146:
```suggestion
        if [[ "$tag_alpha_lt_current_alpha" -eq 1 && "$tag_alpha_gt_candidate_alpha" -eq 1 ]]; then
```

maxdanilov (changes requested):
Looks good overall, leaving some minor suggestions.

previous-version/previous_version.sh
oleksandr-kriuchenko-lohika: @maxdanilov this is a good question and hard for me to answer, as I don't have much experience with those approaches. Here is what AI thinks about it (**TLDR it prefers the current version as more reliable and explicit**):
image
image

maxdanilov: I'd say direct chaining is easier to read and less confusing, so let's go for it.

previous-version/previous_version.sh
oleksandr-kriuchenko-lohika: As far as I understand both curly braces and parentheses are used for command grouping, while the difference is that curly braces execute in same shell, while parentheses start a subshell. Not sure we need to start a subshell here

maxdanilov: Fair enough, let's keep it as is then.

maxdanilov (approved):
Looks good, thanks!

PR #463, opened 2025-08-12
chore: allow retry action and workflow to rerun whole workflow

Langleu:
Hey Infra 👋,

I have a small suggestion: Allow the retry action/workflow to retrigger the whole workflow, not just the failed jobs.

We have a use case where it would be beneficial for us to rerun the whole workflow instead of just the failed jobs. Since we already have the functionality available, I thought to simply extend it instead of recreating it on our side.

For existing users nothing changes as the default is to retry the failed jobs.

Related [gh run rerun docs](https://cli.github.com/manual/gh_run_rerun).

> Rerun an entire run, only failed jobs, or a specific job from a run.

Let me know if names should be changed or implemented differently.
We could also approach it from a different side to say `rerun-failed-jobs` or similar and make it the default to be true and have something like `EXTRA_ARGS` that's passed in that's either empty or `failed` and maybe extensible as extra input to allow passing in anything from outside.

### DIFF
--- a/.github/workflows/retry-worfklow-run.yml
+++ b/.github/workflows/retry-worfklow-run.yml
@@ -38,6 +38,14 @@ on:
           Needed by https://github.com/Codex-/return-dispatch as workaround for https://github.com/cli/cli/issues/4001
         required: false
         default: "not-set"
+      rerun-whole-workflow:
+        description: |
+          If set to true, the workflow will be re-triggered in its entirety.
+          This is useful when you want to retry the entire workflow instead of
+          just failed jobs.
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   retry:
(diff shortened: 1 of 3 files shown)


clementnero (approved):
A nice improvement, thx @Langleu !

> dLet me know if names should be changed or implemented differently.
It's looking good to me.

> We could also approach it from a different side to say rerun-failed-jobs or similar and make it the default to be true and have something like EXTRA_ARGS that's passed in that's either empty or failed and maybe extensible as extra input to allow passing in anything from outside.

I see your point, as it would give more flexibility for users and avoid such (blocking) PRs. In this case the number of extra args are very limited, and adding a new parameter should not happen in the near future (I am also wondering if sanitizing such extra inputs is necessary on a security point of view).

Langleu:
Good point, in that case let's leave it as it is for now and not over-engineer it.
Appreciate the review 🙇 !

PR #501, opened 2025-09-24
feat: Refactoring submit-test-status workflow to also accept files

szpraat:
Updating the submit workflow status to expect and handle `jsonl` file instead of stdout


cmur2:
Please also reflect your changes in the `README.md` documentation for this action, to avoid it becoming outdated.

PR #504, opened 2025-09-29
Refactoring submit-test-status workflow to expect file instead of rea…

szpraat:
### From now on reading from stdout is [deprecated but still works](https://github.com/camunda/camunda/actions/runs/18111485516/job/51538546809?pr=38877#step:11:300):

image

### New solution when [file path is passed to the job](https://github.com/camunda/camunda/actions/runs/18121648912/job/51567645966?pr=38201#step:12:299):
image

### DIFF
--- a/submit-test-status/README.md
+++ b/submit-test-status/README.md
@@ -7,13 +7,21 @@ The GHA uses [Google Big Query batch loading](https://cloud.google.com/bigquery/
 
 This composite GHA can be used in any repository that was set up to provide credentials for accessing the central Google Big Query (BQ) database (credentials via Vault).
 
+**This action supports two input methods:**
+
+1. **File Path Method**: Provide a path to a file containing test results in JSONL format
+2. **Inline Data Method**: Provide test event data directly as a multi-line string
+
 ### Inputs
 
-| Input name           | Description                                        |
-|----------------------|----------------------------------------------------|
-| test_event_record    | Multi-line string that contains the details of the test events in [JSONL format](https://jsonlines.org). One test event per line. |
-| job_name_override    | Optional string being used for the `job_name` field instead of the default `$GITHUB_JOB`, useful e.g. for matrix builds |
-| gcp_credentials_json | Credentials for a Google Cloud ServiceAccount allowed to publish to Big Query formatted as contents of credentials.json file |
+| Input name               | Description                                        | Required |
+|--------------------------|----------------------------------------------------| ---------|
+| test_results_file_path   | Path to a file containing test event(s) in JSONL format. Required keys are test_name and test_status. | No* |
+| test_event_record        | Test event(s) in JSONL format. Required keys are test_name and test_status. | No* |
+| job_name_override        | Optional string being used for the `job_name` field instead of the default `$GITHUB_JOB`, useful e.g. for matrix builds | No |
+| gcp_credentials_json     | Credentials for a Google Cloud ServiceAccount allowed to publish to Big Query formatted as contents of credentials.json file | Yes |
+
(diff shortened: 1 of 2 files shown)


submit-test-status/action.yml
```diff
@@ -33,58 +37,112 @@ runs:
...
+        fi
+
+      elif [ -n "${{ inputs.test_event_record }}" ]; then
+        # Method 2: Create file from inline input
+        INPUT_FILE="$TEMP_DIR/input_test_event_record.jsonl"
+        echo "Using inline data input method, creating: $INPUT_FILE"
+
+        cat <<EOF > $INPUT_FILE
+      ${{ inputs.test_event_record }}
+      EOF
```
cmur2, line 85: 🔧 Indentation here is off, all lines should be indented as the `cat`.

It might not cause any problems like this but if it does, they are probably hard to debug and I'd avoid that.

submit-test-status/action.yml
```diff
@@ -33,58 +37,112 @@ runs:
...
+        echo "SKIP_SUBMISSION=true" >> $GITHUB_ENV
+        exit 0
+      fi
+
+      echo "Final INPUT_FILE: $INPUT_FILE"
+      echo "INPUT_FILE=$INPUT_FILE" >> $GITHUB_ENV
+      echo "SKIP_SUBMISSION=false" >> $GITHUB_ENV
+
+  - name: validate input for file format
+    if: env.SKIP_SUBMISSION != 'true'
```
cmur2, line 105: 🔧 Consider inverting the variable name meaning and usage to avoid double-negation.

When reading this I'm thinking

> So if skipping the submission is NOT true, aka if skipping the submission is false aka if we DONT skip the submission aka if we want to submit

Maybe call it `DO_SUBMIT_METRICS` and use `if: env.DO_SUBMIT_METRICS == 'true'` etc.

submit-test-status/action.yml
```diff
@@ -33,58 +37,112 @@ runs:
...
+
+  - name: ensure that jq is available
+    uses: dcarbone/install-jq-action@v3.2.0
+
+  - name: load input
+    shell: bash
+    env:
+      TEMP_DIR: ${{ runner.temp }}/sts
+    run: |
+      mkdir -p $TEMP_DIR
```
cmur2, line 61: 🔧 `shellcheck` would tell you to put quotes around this variable to prevent word splitting

submit-test-status/README.md
```diff
@@ -7,13 +7,21 @@ The GHA uses [Google Big Query batch loading](https://cloud.google.com/bigquery/
...
-| gcp_credentials_json | Credentials for a Google Cloud ServiceAccount allowed to publish to Big Query formatted as contents of credentials.json file |
+| Input name               | Description                                        | Required |
+|--------------------------|----------------------------------------------------| ---------|
+| test_results_file_path   | Path to a file containing test event(s) in JSONL format. Required keys are test_name and test_status. | No* |
+| test_event_record        | Test event(s) in JSONL format. Required keys are test_name and test_status. | No* |
+| job_name_override        | Optional string being used for the `job_name` field instead of the default `$GITHUB_JOB`, useful e.g. for matrix builds | No |
+| gcp_credentials_json     | Credentials for a Google Cloud ServiceAccount allowed to publish to Big Query formatted as contents of credentials.json file | Yes |
+
+*Note: Either `test_results_file_path` OR `test_event_record` must be provided (not both). If neither is provided, the action will skip submission gracefully.
```
cmur2, line 24: ✨ Really nice way to capture the requirement, easy to follow 👍🏼

cmur2 (approved):

PR #680, opened 2026-05-07
chore(submit-aborted-gha-status): massive speedup due to using REST API instead of gcloud CLI

cmur2:
## Description

This PR is a direct transplantation of https://github.com/camunda/infra-global-github-actions/pull/600 to another action: `submit-aborted-gha-status`.

Impact: the Unified CI in the Monorepo has a `check-results` job that calls this action once for each build, thus prolonging the feedback cycle by ~10 seconds unnecessarily.

## 🚀 Performance Improvement

This PR delivers a **massive performance improvement** to the `submit-aborted-gha-status` action by replacing the slow Google Cloud SDK installation with direct BigQuery REST API calls. The `setup-gcloud` action installs the entire Cloud SDK package, which is unnecessary when we only need to insert a few rows into BigQuery.

## 📊 Performance Impact

**Before:** ~10 seconds (SDK installation overhead)  
**After:** ~2 seconds (direct API call)  
**Improvement:** ⚡ **~5x faster** ⚡

In April 2026 there were ~50k `check-results` job runs (one per Unified CI workflow run): `SELECT COUNT(DISTINCT SPLIT(build_id, '/')[0]) as number_of_all_workflow_runs FROM ci-30-162810.prod_ci_analytics.build_status_v2 WHERE report_time >= '2026-04-01' AND report_time < '2026-04-30'`

Let's assume each of those submissions would be 10s faster than before (conservative).

The overall **GHA build minute savings in theory can be as high as ~8000 minutes, or ~130 hours per month.** (We don't pay for `ubuntu-latest` in `camunda/camunda` but it quickens the feedback cycle still.)

### Removed (Slow):
- ❌ `google-github-actions/setup-gcloud@v3` - Installs entire gcloud SDK (~20-30 seconds)
- ❌ `gcloud info` - Version check step
- ❌ `bq` CLI tool - Command-line BigQuery client

### Added (Fast):
- ✅ `token_format: 'access_token'` - Export OAuth token from auth action
- ✅ Direct BigQuery REST API call using `curl` - Native HTTP request
- ✅ JSON payload construction in bash - No external dependencies

## 🛠️ Technical Details

The action now:
1. Uses `google-github-actions/auth@v3` to authenticate (already in use)
2. Requests an access token output from the auth action, need https://github.com/camunda/infra-core/pull/11525
3. Constructs BigQuery `tabledata.insertAll` API payload directly
4. Makes a direct HTTPS POST request to BigQuery REST API using `curl` in a loop (in practice there are only a few iterations, so not investing into batching)

All functionality remains identical - same inputs, same data format, same BigQuery table structure.

## ✅ Benefits

- **Faster CI/CD pipelines** - Reduces overhead in every job that reports build status
- **Lower resource usage** - No SDK download/installation required
- **Simpler dependencies** - Uses `curl` which is pre-installed on all runners
- **Same reliability** - Uses official Google Cloud REST APIs
- **Cross-platform** - Works on Linux, macOS, and Windows runners

## 🔍 Testing

https://github.com/camunda/camunda/pull/52686

The action maintains full backward compatibility:
- ✅ All existing inputs work identically
- ✅ Same BigQuery table schema
- ✅ Same authentication mechanism (service account JSON)
- ✅ Same error handling behavior

### DIFF
--- a/submit-aborted-gha-status/action.yml
+++ b/submit-aborted-gha-status/action.yml
@@ -30,23 +30,23 @@ runs:
     uses: google-github-actions/auth@v3
     with:
       credentials_json: '${{ inputs.gcp_credentials_json }}'
-
-  - name: Setup Google Cloud SDK
-    uses: google-github-actions/setup-gcloud@v3
-
-  - name: Print Google Cloud SDK version used
-    shell: bash
-    run: |
-      gcloud info
+      token_format: 'access_token'
 
   - name: Submit build status to CI Analytics
     shell: bash
     env:
-      BG_COMMAND: "${{ (runner.os == 'Windows') && 'bq.cmd' || 'bq' }}"
       BUILD_BASE_REF: "${{ github.base_ref && format('refs/heads/{0}', github.base_ref) || github.event.merge_group.base_ref }}"
       BUILD_HEAD_REF: "${{ github.head_ref && format('refs/heads/{0}', github.head_ref) || github.event.merge_group.head_ref }}"
       GH_TOKEN: ${{ github.token }}
+      ACCESS_TOKEN: "${{ steps.auth.outputs.access_token }}"
     run: |
+      # ── Parse BigQuery table name (format: project_id:dataset.table) ──
+      TABLE_NAME="${{ inputs.big_query_table_name }}"
+      PROJECT_ID="${TABLE_NAME%%:*}"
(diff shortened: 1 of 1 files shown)


kellervater (approved):
looks good to me. this is an awesome improvement. also on 600.
However, you might want to check the remark from Copilot.

PR #702, opened 2026-06-02
feat(vault): add OIDC/JWT auth support to vault composites

clementnero:
## Description

Related to https://github.com/camunda/team-infrastructure/issues/1035.

Adds GitHub OIDC/JWT authentication support to two Vault composite actions, alongside the existing AppRole method:

- `generate-github-app-token-from-vault-secrets`
- `rerun-failed-run`

This unblocks consumers migrating Vault auth from AppRole to GitHub OIDC/JWT.

Changes:
- New optional inputs for JWT auth (`*-jwt-path`, `*-jwt-role`, `*-jwt-audience`)
- Existing AppRole inputs made optional (kept working) → **backward compatible**
- `rerun-failed-run` gains a `vault-auth-method` input (defaults to `approle`)
- READMEs updated with JWT usage examples (note: callers must grant `id-token: write`)

### DIFF
--- a/generate-github-app-token-from-vault-secrets/README.md
+++ b/generate-github-app-token-from-vault-secrets/README.md
@@ -13,15 +13,20 @@ This composite GHA can be used in any repository.
 | github-app-id-vault-path            | The path of the Vault secret storing the ID of the GitHub App           |
 | github-app-private-key-vault-key    | The key of the Vault secret storing the private key of the GitHub App   |
 | github-app-private-key-vault-path   | The path of the Vault secret storing the private key of the GitHub App  |
-| vault-auth-method                   | The method to use to authenticate with Vault (*)                        |
-| vault-auth-role-id                  | The Role Id for (Vault) App Role authentication                         |
-| vault-auth-secret-id                | The Secret Id for (Vault) App Role authentication                       |
+| vault-auth-method                   | The method to use to authenticate with Vault: `approle` or `jwt` (*)    |
+| vault-auth-role-id                  | The Role Id for (Vault) App Role authentication (required when `vault-auth-method=approle`) |
+| vault-auth-secret-id                | The Secret Id for (Vault) App Role authentication (required when `vault-auth-method=approle`) |
+| vault-auth-jwt-path                 | The auth backend path for (Vault) JWT authentication (required when `vault-auth-method=jwt`) |
+| vault-auth-jwt-role                 | The role for (Vault) JWT authentication (required when `vault-auth-method=jwt`) |
+| vault-auth-jwt-audience             | The GitHub OIDC audience for (Vault) JWT authentication (required when `vault-auth-method=jwt`) |
 | vault-url                           | The URL for the Vault endpoint                                          |
 | skip-token-revoke                   | If truthy, the token will not be revoked when the current job is complete (optional) |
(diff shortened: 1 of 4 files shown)


generate-github-app-token-from-vault-secrets/action.yml
```diff
@@ -73,6 +121,9 @@ runs:
       method: ${{ inputs.vault-auth-method }}
       roleId: ${{ inputs.vault-auth-role-id }}
       secretId: ${{ inputs.vault-auth-secret-id }}
+      path: ${{ inputs.vault-auth-jwt-path }}
+      role: ${{ inputs.vault-auth-jwt-role }}
+      jwtGithubAudience: ${{ inputs.vault-auth-jwt-audience }}
       exportEnv: false
```
kellervater, line 137: does this action gracefully handle if we provide both, approle/secret and jwt values?
if so, what takes precedence?

generate-github-app-token-from-vault-secrets/README.md
kellervater, line 1: we maybe should discourage approle/secret usage in here as long as we provide both.

rerun-failed-run/action.yml
```diff
@@ -47,14 +65,48 @@ inputs:
...
+          *)
+            echo "::error::Unsupported vault-auth-method '${VAULT_AUTH_METHOD}' (expected 'approle' or 'jwt')"
+            exit 1
+            ;;
+        esac
+        if [[ ${#missing[@]} -gt 0 ]]; then
+          echo "::error::Missing required input(s) for vault-auth-method='${VAULT_AUTH_METHOD}': ${missing[*]}"
+          exit 1
+        fi
+
```
kellervater, line 108: since this is the 2nd time I spot this snippet: do you think it makes sense to extract this into a common shell script (only in case it's easy to do as it's an action and path handling might need extra checkouts)

rerun-failed-run/README.md
kellervater, line 1: maybe also here: Discouraging users from using approle might be useful.

kellervater (approved):
nice. I like the backwards compatibility!
all comments are non-blocking.

rerun-failed-run/action.yml
clementnero: I see your point, but I don’t see a need right now to refactor this into a single script, since this is meant to be temporary and will be removed once all workflows are migrated to the JWT auth method. Also each action might take a different path with different additional checks...

generate-github-app-token-from-vault-secrets/README.md
clementnero: The full migration will be handled by a separate issue. For consistency, such warning messages will be added as part of the later issue and dedicated announcements.

rerun-failed-run/README.md
clementnero: Same as https://github.com/camunda/infra-global-github-actions/pull/702#discussion_r3347608569

generate-github-app-token-from-vault-secrets/action.yml
clementnero: Hmm, good point. I’ll update the check step to enforce using one set of params or the other.

rerun-failed-run/action.yml
kellervater: fair point

PR #740, opened 2026-07-08
docs: document fallback-base-ref and ancestor-staleness filter in dependency-vuln-check

szpraat:
## What

Documents two behaviours introduced in #739 that were missing from the README.

### Changes

**1. `fallback-base-ref` input**

Added to the inputs table and the usage example. This input was shipped in #739 but never documented.

**2. Fail-closed table — stacked PR fallback**

The previous table said "no snapshotted ancestor → fail closed", which is no longer accurate. When `base-ref` has no snapshots and differs from `fallback-base-ref`, the gate retries on the fallback branch before failing closed. Updated the row to reflect this.

**3. Ancestor-staleness filter (Pattern A)**

New subsection under "Base-snapshot resolution & fail-closed contract" explaining the `effective_base...latest_on_base_branch` drift filter: dep versions bumped on the base branch after the snapshot was taken are moved to informational rather than blocking, preventing false positives on unrelated Renovate bumps.

## Related

- #739 — the PR that introduced these behaviours

### DIFF
--- a/dependency-vuln-check/README.md
+++ b/dependency-vuln-check/README.md
@@ -37,6 +37,7 @@ version shows up as `added` in the diff and is evaluated normally.
     base-ref: ${{ github.event.pull_request.base.ref }}
     snapshot-workflow: maven-dependency-snapshot.yml
     # optional — defaults shown
+    fallback-base-ref: main
     max-snapshot-lookback: "30"
     override-label: ci:vuln-gate-override
     config-file: .github/dependency-review-config.json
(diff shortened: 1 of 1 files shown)


maxdanilov (approved):

szpraat:
Superseded by #741, which now carries these README updates alongside the actual fix — and extends them to the union (snapshot-tip + base-sha) ancestor-staleness filter. Closing in favor of #741.

PR #741, opened 2026-07-08
fix: key pre-existing dep filter on base-sha, not latest snapshot

szpraat:
## Problem

The pre-existing-dependency filter (`_base_branch_pre_existing`, added in #739) compared `effective_base ... latest_snapshot_commit` to find deps the base branch introduced after the snapshot.

Natively-detected ecosystems (Go modules, npm) **never trigger the Maven snapshot workflow** (`maven-dependency-snapshot.yml` is path-filtered to poms). So after such a dep is bumped on `main`, the latest snapshot commit stays equal to `effective_base` — the filter's precondition short-circuits, the drift window collapses to nothing, and the false positive slips through.

### Observed incident (camunda/camunda, 2026-07-08)

`golang.org/x/crypto@0.51.0` (9 advisories, critical/high, no fix) in `load-tests/metrics-exporter/go.mod` falsely blocked **3 Renovate PRs that never touched Go files**:

| PR | Update | Ecosystem |
|----|--------|-----------|
| camunda/camunda#57110 | react-hook-form 7.81.0 | JS |
| camunda/camunda#57054 | langchain4j-bom 1.17.1 | Java |
| camunda/camunda#56937 | jnr-ffi 2.3.0 | Java |

`crypto@0.51.0` was introduced on `main` by an unrelated testcontainers bump **after** the effective-base snapshot — none of the three PRs added it.

## Fix

Diff `effective_base` against **two** reference points and union the `added` triples. The references have complementary coverage:

| Reference | Maven drift | Go/npm drift |
|-----------|:-----------:|:------------:|
| latest snapshot commit (`latest_on_branch`) | ✅ | ❌ (window collapses) |
| `base_sha` (true branch tip) | ❌ (not snapshotted) | ✅ |

Maven deps exist in the dependency graph **only at snapshotted commits**, so the snapshot tip is the only reference that surfaces Maven base-branch drift (the case #739 fixed — e.g. jackson-databind). `base_sha` is usually not a snapshot commit, so Maven deps there show as `removed` and escape the filter — but it reflects the true tip for natively-detected ecosystems. **Neither alone is sufficient; the union covers both.**

- No extra work beyond one more compare — `base_sha` is already resolved in `main()`.
- No coverage loss — genuinely PR-introduced vulns still block (including shipped Go like `c8run`).
- **Per-reference fail-open**: if one compare fails, only that reference is skipped; the other still applies. Real vulns are never suppressed by a failure.

## Tests

`57 passed`. New tests cover: union of both references, dedupe when the two tips are equal, per-reference partial failure, all-references failure (fail-open), and a `main()` regression driving the exact Go false-positive scenario (asserts both the snapshot tip and base_sha are passed).

## Docs

This PR also updates the `dependency-vuln-check` README (folded in from the now-closed #740): documents the `fallback-base-ref` input, the stacked-PR row in the fail-closed contract, and the section **"Ignoring dependencies the base branch itself added"** — the filter that stops a PR being blamed for a dependency `main` introduced — described with the union (latest-snapshot + base-sha) behaviour introduced here.

## Not a revert of #739

All of #739's filter machinery is retained (`_base_branch_pre_existing`, the `pre_existing` bucket, the PR-comment section, the stacked-PR fallback). This only changes *which* reference commits the filter diffs against.

## Related

- #739 — introduced the pre-existing-dependency filter this hardens
- #740 — README docs originally split out; now **closed and folded into this PR** (see Docs above)

### DIFF
--- a/dependency-vuln-check/README.md
+++ b/dependency-vuln-check/README.md
@@ -37,6 +37,7 @@ version shows up as `added` in the diff and is evaluated normally.
     base-ref: ${{ github.event.pull_request.base.ref }}
     snapshot-workflow: maven-dependency-snapshot.yml
     # optional — defaults shown
+    fallback-base-ref: main
     max-snapshot-lookback: "30"
     override-label: ci:vuln-gate-override
     config-file: .github/dependency-review-config.json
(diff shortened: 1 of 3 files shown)


cmur2 (approved):
LGTM

@szpraat at some point when the mechanism is stable, you should do a short presentation in EPOD meeting about the approach and high-level flow. This logic got quite complicated quite quickly.

PR #751, opened 2026-07-17
chore: include the build-duration-milliseconds github runner monitoring in the reusable workflow

Kerruba:
## Description

Centralize build-duration collection in the shared submit-build-status action

Previously repositories were calculating build_duration_millis in its own observe-build-status wrapper (duplicated logic, relying on $GITHUB_ACTION_PATH directory mtime). This moves the calculation into the shared start-build-monitor / submit-build-status actions.

**What changed:**

- start-build-monitor now records the job start time with millisecond precision (date +%s%3N) into a shared temp file, and clears any stale monitor state (PID/log/timestamp) at startup. It must remain the first job step so the duration covers the
whole job.
- submit-build-status reads that timestamp when it stops the monitor, computes the elapsed duration, validates it (non-negative integer, capped at 72h / 259200000 ms), and submits it automatically. Invalid/missing/out-of-range timestamps omit the
field with a warning instead of failing submission.
- Duration and resource/network metrics are now collected before Google auth and the BigQuery submission, so they exclude telemetry-submission overhead.
- Resolution precedence: explicit build_duration_millis input → monitor-derived duration → field omitted.
- Resource-metric computation (CPU/mem percentiles, network bytes) is unchanged.

**Backward compatibility: Fully backward compatible.**

- The optional build_duration_millis input is retained. Repositories that pass their own value override the automatic duration — an explicit non-empty input always takes precedence, so unmigrated wrappers keep working unchanged.
- Repositories that do not run start-build-monitor (or haven't migrated) simply get no automatic duration; the field is omitted (NULL in BigQuery), exactly as before.
- No consumer changes are required to adopt this; wrappers can drop their local duration calculation later, after this action ref is available.

### DIFF
--- a/start-build-monitor/README.md
+++ b/start-build-monitor/README.md
@@ -1,6 +1,6 @@
 # start-build-monitor
 
-This composite GitHub Action starts a lightweight background CPU and memory monitor at the beginning of a job. When used together with [`submit-build-status`](../submit-build-status/), resource utilization metrics are automatically included at the end in the [CI Analytics](https://confluence.camunda.com/display/HAN/CI+Analytics) record for that job.
+This composite GitHub Action starts a lightweight background CPU and memory monitor at the beginning of a job. When used together with [`submit-build-status`](../submit-build-status/), resource utilization metrics and the job duration are automatically included at the end in the [CI Analytics](https://confluence.camunda.com/display/HAN/CI+Analytics) record for that job.
 
 ## Usage
 
@@ -25,11 +25,11 @@ jobs:
           gcp_credentials_json: "${{ secrets.YOUR_GCP_CREDENTIALS }}"
 ```
 
-Jobs that do **not** include `start-build-monitor` are unaffected — `submit-build-status` simply omits the resource fields, which appear as `NULL` in BigQuery like any other optional field.
+Jobs that do **not** include `start-build-monitor` are unaffected — `submit-build-status` simply omits the resource and duration fields, which appear as `NULL` in BigQuery like any other optional field.
 
 ## Behavior
 
-Starts a polling loop in the background (interval: 5s) that samples CPU and memory usage once per interval until `submit-build-status` stops it at job end.
+Starts a polling loop in the background (interval: 5s) that samples CPU and memory usage once per interval until `submit-build-status` stops it at job end. It also records the start time with millisecond precision. `submit-build-status` uses this timestamp to calculate the duration, from this action's start until resource collection begins. Keep this action as the **first job step** so the reported duration covers the entire job.
 
 ### Metric sources
 
--- a/start-build-monitor/start-build-monitor.sh
+++ b/start-build-monitor/start-build-monitor.sh
@@ -2,8 +2,13 @@
 
 INTERVAL=5  # seconds
 # NOTE: these paths are shared with the monitor stop script(s) — if changed, update both
-LOG_FILE=/tmp/_monitor-start.log
-PID_FILE=/tmp/_monitor-start.pid
+LOG_FILE="${BUILD_MONITOR_LOG_FILE:-/tmp/_monitor-start.log}"
+PID_FILE="${BUILD_MONITOR_PID_FILE:-/tmp/_monitor-start.pid}"
+START_TIME_FILE="${BUILD_MONITOR_START_TIME_FILE:-/tmp/_monitor-start.epoch-millis}"
+
+# Remove state from a previous job before establishing this job's start time.
+rm -f "$PID_FILE" "$LOG_FILE" "$START_TIME_FILE"
+date +%s%3N > "$START_TIME_FILE"
 
 # CPU and memory cgroup controllers are independently enabled, so detect them separately.
 # Cgroup-based metrics are scoped to this container — unaffected by other pods on the node.
@@ -103,4 +108,4 @@ INTERVAL_US=$(( INTERVAL * 1000000 ))
 ) >> "$LOG_FILE" 2>&1 &
 
 echo $! > "$PID_FILE"
-echo "Resource monitor started (PID=$(cat $PID_FILE), cpu=${CPU_SOURCE}(${CPU_LIMIT_CORES} cores) mem=${MEM_SOURCE}(${MEM_TOTAL_KB}KB), interval=${INTERVAL}s)"
+echo "Resource monitor started (PID=$(cat "$PID_FILE"), cpu=${CPU_SOURCE}(${CPU_LIMIT_CORES} cores) mem=${MEM_SOURCE}(${MEM_TOTAL_KB}KB), interval=${INTERVAL}s)"
--- a/submit-build-status/action.yml
+++ b/submit-build-status/action.yml
@@ -8,7 +8,7 @@ inputs:
     description: String representing the build status that should be submitted to CI Analytics, e.g. "success", "failed", "cancelled".
     required: true
   build_duration_millis:
-    description: Optional number (positive) that indicates the duration of the build in milliseconds.
+    description: Optional non-negative build duration in milliseconds. Overrides the duration collected by start-build-monitor.
     required: false
   user_reason:
     description: Optional string (200 chars max) the user can submit to indicate the reason why a build has ended with a certain build status , e.g. "flaky-tests".
@@ -36,22 +36,12 @@ inputs:
 runs:
   using: composite
   steps:
-  - name: Echo inputs
-    shell: bash
-    run: |
-      echo "Inputs"
-      echo "-----"
-      echo "Build status: ${{ inputs.build_status }}"
-      echo "Build duration (in ms): ${{ inputs.build_duration_millis }}"
-      echo "User reason: ${{ inputs.user_reason }}"
-      echo "User description: ${{ inputs.user_description }}"
-      echo "Job name override: ${{ inputs.job_name_override }} (default job name: $GITHUB_JOB)"
-      echo "BQ table name (for testing): ${{ inputs.big_query_table_name }}"
-
   - name: Validate inputs
+    id: validate
     shell: bash
     env:
       BUILD_STATUS: "${{ inputs.build_status }}"
+      BUILD_DURATION_MILLIS: "${{ inputs.build_duration_millis }}"
       USER_REASON: "${{ inputs.user_reason }}"
       USER_DESCRIPTION: "${{ inputs.user_description }}"
     run: |
@@ -69,6 +59,40 @@ runs:
         echo "Specified user description has more than 1000 characters!"
         exit 1
       fi
+      VALIDATED_BUILD_DURATION_MILLIS=$(bash "$GITHUB_ACTION_PATH/validate-build-duration.sh" "$BUILD_DURATION_MILLIS")
+      printf 'build_duration_millis=%s\n' "$VALIDATED_BUILD_DURATION_MILLIS" >> "$GITHUB_OUTPUT"
+
+  - name: Collect resource usage metrics
+    id: monitor
+    if: always()
+    shell: bash
+    run: |
+      bash "$GITHUB_ACTION_PATH/stop-and-collect-build-monitor.sh" "$GITHUB_OUTPUT"
+
+  - name: Resolve build duration
+    id: duration
+    shell: bash
+    env:
+      EXPLICIT_BUILD_DURATION_MILLIS: "${{ steps.validate.outputs.build_duration_millis }}"
+      MONITOR_BUILD_DURATION_MILLIS: "${{ steps.monitor.outputs.build_duration_millis }}"
+    run: |
+      BUILD_DURATION_MILLIS=$(bash "$GITHUB_ACTION_PATH/resolve-build-duration.sh" \
+        "$EXPLICIT_BUILD_DURATION_MILLIS" "$MONITOR_BUILD_DURATION_MILLIS")
+      printf 'build_duration_millis=%s\n' "$BUILD_DURATION_MILLIS" >> "$GITHUB_OUTPUT"
+
+  - name: Echo inputs
+    shell: bash
+    env:
+      BUILD_DURATION_MILLIS: "${{ steps.duration.outputs.build_duration_millis }}"
+    run: |
+      echo "Inputs"
+      echo "-----"
+      echo "Build status: ${{ inputs.build_status }}"
+      echo "Build duration (in ms): $BUILD_DURATION_MILLIS"
+      echo "User reason: ${{ inputs.user_reason }}"
+      echo "User description: ${{ inputs.user_description }}"
+      echo "Job name override: ${{ inputs.job_name_override }} (default job name: $GITHUB_JOB)"
+      echo "BQ table name (for testing): ${{ inputs.big_query_table_name }}"
 
   - name: Login to Google Cloud
     id: auth
@@ -77,22 +101,21 @@ runs:
       credentials_json: '${{ inputs.gcp_credentials_json }}'
       token_format: 'access_token'
 
-  - name: Collect resource usage metrics
-    id: monitor
-    shell: bash
-    run: |
-      bash "$GITHUB_ACTION_PATH/stop-and-collect-build-monitor.sh" "$GITHUB_OUTPUT"
-
   - name: Submit build status to CI Analytics
     shell: bash
     env:
       BUILD_BASE_REF: "${{ github.base_ref && format('refs/heads/{0}', github.base_ref) || github.event.merge_group.base_ref }}"
       BUILD_HEAD_REF: "${{ github.head_ref && format('refs/heads/{0}', github.head_ref) || github.event.merge_group.head_ref }}"
+      BUILD_DURATION_MILLIS: "${{ steps.duration.outputs.build_duration_millis }}"
       ACCESS_TOKEN: "${{ steps.auth.outputs.access_token }}"
       RUNNER_INFO_RUNS_ON_FILE: "/home/runner/.camunda-arc-runner-info/runs-on"
     run: |
       # ── Build monitor: get resource usage metrics (might be empty) ──
       MONITOR_FIELDS="${{ steps.monitor.outputs.monitor_fields }}"
+      BUILD_DURATION_FIELD=""
+      if [ -n "$BUILD_DURATION_MILLIS" ]; then
+        BUILD_DURATION_FIELD=$(printf ', "build_duration_milliseconds": "%s"' "$BUILD_DURATION_MILLIS")
+      fi
 
       # ── Read runs-on label info from self-hosted runner metadata (might be empty) ──
       RUNNER_TYPE_FIELD=""
@@ -129,7 +152,7 @@ runs:
         "runner_os": "$(echo $RUNNER_OS | tr '[:upper:]' '[:lower:]')"
         ${{ (env.BUILD_BASE_REF == '') && ' ' || format(', "build_base_ref": "{0}"', env.BUILD_BASE_REF) }}
         ${{ (env.BUILD_HEAD_REF == '') && ' ' || format(', "build_head_ref": "{0}"', env.BUILD_HEAD_REF) }}
-        ${{ (inputs.build_duration_millis == '') && ' ' || format(', "build_duration_milliseconds": "{0}"', inputs.build_duration_millis) }}
+        $BUILD_DURATION_FIELD
         ${{ (inputs.user_reason == '') && ' ' || format(', "user_reason": "{0}"', inputs.user_reason) }}
         ${{ (inputs.user_description == '') && ' ' || format(', "user_description": "{0}"', inputs.user_description) }}
         $MONITOR_FIELDS
--- a/submit-build-status/stop-and-collect-build-monitor.sh
+++ b/submit-build-status/stop-and-collect-build-monitor.sh
@@ -8,78 +8,111 @@ fi
 
 MONITOR_FIELDS=""
 # NOTE: these paths are shared with start-build-monitor.sh -- if changed, update both
-PID_FILE=/tmp/_monitor-start.pid
-LOG_FILE=/tmp/_monitor-start.log
+PID_FILE="${BUILD_MONITOR_PID_FILE:-/tmp/_monitor-start.pid}"
+LOG_FILE="${BUILD_MONITOR_LOG_FILE:-/tmp/_monitor-start.log}"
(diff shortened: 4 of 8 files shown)


cmur2:
High-level feedback:

> relying on an unreliable GITHUB_ACTION_PATH directory mtime

Bold claim that it is unreliable, it worked so far 😉

> start-build-monitor now records the job start time
> :
> Resolution precedence: explicit build_duration_millis input → monitor-derived duration → field omitted.

Does that mean that Monorepo CI jobs retain their original behavior and don't need a massive refactor to make them all call `start-build-monitor`? Asking because while we want duration information, we don't care about performance metrics for many jobs.

That said, did you explore other (implicit) ways of deriving the job/runner start time? From the user's perspective what mostly matters is how long they have to wait for CI feedback.

> start-build-monitor now records the job start time with millisecond precision (date +%s%3N) into a shared temp file, and clears any stale monitor state (PID/log/timestamp) at startup. It must remain the first job step so the duration covers the
whole job.

Two problems I see here:

1. `start-build-monitor` gets started later than cloning the Git repository & composite GHAs from other repos (which is what `GITHUB_ACTION_PATH` represents) so actually the job duration reported now will be consistenly lower than before - we have to be aware of & fine with this systematic behavior change
    * Both approaches are only estimating the real duration.
2. Given the 1st problem, I doubt that using "millisecond precision" will give us tangible benefits in regards to the data. My original approach intentionally capped the precision at seconds to avoid suggesting that there is more precision available than the measurement tooling actually provides.

> Duration and resource/network metrics are now collected before Google auth and the BigQuery submission, so they exclude telemetry-submission overhead.

How long do Google auth and the BigQuery submission usually take? I think most jobs that do not submit test results complete this within seconds. Asking because users overall care about how long their CI feedback takes, not what the runner does, so the total runtime seems like a more representative value.

cmur2:
Luca and I just had a call, and as outcome I want to document some context:

* the current duration measurement method employed `observe-build-status` composite action in the monorepo is a bit of a hack, relying on implicit assumptions (that Git clone happens early and first)
* it was chosen as an approximation delivering "good enough" measurements that are accurate in the order of multiple seconds/1 minute (the real error is probably underestimating job duration by several seconds)
* for this PR I mainly want to ensure two things from Monorepo CI perspective:
    1. avoid necessity to refactor many/all GHA jobs in `camunda/camunda` as consequence of this PR
    2. ensure measurements stay "compatible" aka include the same measurement error as before this PR

Both 1. and 2. (partly) should be addressed with this fallback logic that Luca provided:

> Resolution precedence: explicit build_duration_millis input → monitor-derived duration → field omitted.

What could disrupt 2. is if this change is also in effect when using explicit `build_duration_millis` input:

> exclude telemetry-submission overhead

Kerruba:
> Bold claim that it is unreliable, it worked so far 😉

I admit I let Claude prepare the PR description here without checking too much about what he said 😓  

> Does that mean that Monorepo CI jobs retain their original behavior and don't need a massive refactor to make them all call start-build-monitor?

The action was developed to support the existing build-time duration calculation. For this reason, this should require no additional activities for repositories that already calculate the job duration in their own way.
I've already tested this workflow in `infra-core`, where the job duration was calculated the same way as the Monorepo.

>  Asking because while we want duration information, we don't care about performance metrics for many jobs.

Performance metrics will be used to support runner right-sizing as defined in camunda/team-infrastructure#1051
I suppose we can always think of adding a way to enable/disable metrics collection on demand, but in the future.

>  That said, did you explore other (implicit) ways of deriving the job/runner start time? 

No, I didn't explore any other way of deriving the runner's start time. I was trying to support job duration centralization while keeping in mind the issue with `GITHUB_ACTION_PATH` not being available in reusable workflows. I can try to think of alternatives here in case.

**UPDATE:** An option for calculating the overall workflow duration could be to find a file in the runner that's created when the runner was created and use that as an estimate of the workflow duration.

> From the user's perspective, what matters most is how long they have to wait for CI feedback.

From an infra/FinOps perspective, job duration will serve as a metric for comparing GitHub action providers. 
From the explorer generated out of the GitHub Enterprise Detailed Report - https://blog.int.camunda.com/posts/2026-07-01-updated-ci-explorer/index.html - we have a breakdown of the GitHub runners' time, but we are currently lacking a breakdown of minutes used on the self-hosted runners. 
This information could be very useful to better estimate the cost and evaluate alternatives to reduce the overall price we pay for the runners

> start-build-monitor gets started later than cloning the Git repository & composite GHAs from other repos (which is what GITHUB_ACTION_PATH represents), so actually the job duration reported now will be consistently lower than before - we have to be aware of & fine with this systematic behavior change
Both approaches only estimate the actual duration.

Ok, but I'm a bit confused here. From the build_status_v2 schema, having "workflow_name" and "job_name" paired with duration_in_milliseconds suggests we are reporting the job duration, not the time elapsed between code checkout and submission to "build_status_v2". The assumption I had until now was that we were monitoring specific job durations/performance, but that's not always true, as it depends on how far in the past the code was checked out.
If we aim to monitor specific job metrics (duration, etc.). I would think the approach above would provide more precise estimates. WDYT?

**UPDATE:** After discussing this with @cmur2, I realized that each repository is able to define how the job duration is calculated. In the Monorepo, for example, the job duration is calculated from the code checkout. To avoid breaking existing metrics and monitoring systems, **the reusable workflow should maintain the ability to support custom-built duration calculations while providing new defaults when no duration is provided**. 
**NOTE:** A potential issue I could see with this "fallback" approach is when, in a repository, the two approaches are mixed (e.g., old workflows use existing calculation, new workflow adopts the fallback calculation).  This could lead to inconsistencies and potential issues/misunderstanding. To avoid this completely, I'm afraid the "fallback" should be provided (and therefore this entire PR should be dropped). WDYT @cmur2 and @clementnero?

> Given the 1st problem, I doubt that using "millisecond precision" will give us tangible benefits regarding the data

I got misled by the composite action "Get build duration in milliseconds", so I guess it's ok to calculate seconds and readopt the same approach of *1000.

cmur2:
> To avoid breaking existing metrics and monitoring systems, the reusable workflow should maintain the ability to support custom-built duration calculations while providing new defaults when no duration is provided.

Agree, this is a fair tradeoff and we offer a clear path forward.

> potential issue I could see with this "fallback" approach is when, in a repository, the two approaches are mixed (e.g., old workflows use existing calculation, new workflow adopts the fallback calculation). This could lead to inconsistencies and potential issues/misunderstanding.

That problem should then be owned by the repository owners, I don't think it should be a blocker.

start-build-monitor/README.md
```diff
@@ -25,11 +25,11 @@ jobs:
           gcp_credentials_json: "${{ secrets.YOUR_GCP_CREDENTIALS }}"
 ```
 
-Jobs that do **not** include `start-build-monitor` are unaffected — `submit-build-status` simply omits the resource fields, which appear as `NULL` in BigQuery like any other optional field.
+Jobs that do **not** include `start-build-monitor` are unaffected — `submit-build-status` simply omits the resource and duration fields, which appear as `NULL` in BigQuery like any other optional field.
 
 ## Behavior
 
-Starts a polling loop in the background (interval: 5s) that samples CPU and memory usage once per interval until `submit-build-status` stops it at job end.
+Starts a polling loop in the background (interval: 5s) that samples CPU and memory usage once per interval until `submit-build-status` stops it at job end. It also records the start time with millisecond precision. `submit-build-status` uses this timestamp to calculate the duration, from this action's start until resource collection begins. Keep this action as the **first job step** so the reported duration covers the entire job.
```
cmur2, line 32: > so the reported duration covers the entire job

💭 ...and also CPU and memory sampling covers the whole job - basically, just always keep it first for any measuring :D

start-build-monitor/start-build-monitor.sh
```diff
@@ -2,8 +2,13 @@
...
 INTERVAL=5  # seconds
 # NOTE: these paths are shared with the monitor stop script(s) — if changed, update both
-LOG_FILE=/tmp/_monitor-start.log
-PID_FILE=/tmp/_monitor-start.pid
+LOG_FILE="${BUILD_MONITOR_LOG_FILE:-/tmp/_monitor-start.log}"
+PID_FILE="${BUILD_MONITOR_PID_FILE:-/tmp/_monitor-start.pid}"
+START_TIME_FILE="${BUILD_MONITOR_START_TIME_FILE:-/tmp/_monitor-start.epoch-millis}"
+
+# Remove state from a previous job before establishing this job's start time.
+rm -f "$PID_FILE" "$LOG_FILE" "$START_TIME_FILE"
```
cmur2, line 10: 💭 We always use clean runners and assume so in several places inside the CI, but no problem making extra sure the files are absent.

submit-build-status/action.yml
```diff
@@ -36,22 +36,12 @@ inputs:
...
-    run: |
-      echo "Inputs"
-      echo "-----"
-      echo "Build status: ${{ inputs.build_status }}"
-      echo "Build duration (in ms): ${{ inputs.build_duration_millis }}"
-      echo "User reason: ${{ inputs.user_reason }}"
-      echo "User description: ${{ inputs.user_description }}"
-      echo "Job name override: ${{ inputs.job_name_override }} (default job name: $GITHUB_JOB)"
-      echo "BQ table name (for testing): ${{ inputs.big_query_table_name }}"
-
```
cmur2, line 50: 🔧 We cannot move this part, input echoing should always happen first to aid debugging. Especially since input validation (next step) might already abort the run.

I assume you move it to save some time from the duration measurement, but a couple of `echo` statements won't delay the job significantly anyways.

submit-build-status/test_build_duration.sh
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env bash
```
cmur2, line 1: ❓ Do we want to commit test harnesses that are not exercised by CI? So far we chose not to.

🔧 _If_ we want to keep it, the file name is misleading since it tests much more than just "build duration" (namely also debug output format for CPU and stuff).

submit-build-status/validate-build-duration.sh
```diff
@@ -0,0 +1,24 @@
...
+if ! [[ "$BUILD_DURATION_MILLIS" =~ ^[0-9]+$ ]]; then
+  echo "Build duration must be a non-negative integer." >&2
+  exit 1
+fi
+
+NORMALIZED_BUILD_DURATION_MILLIS="${BUILD_DURATION_MILLIS#"${BUILD_DURATION_MILLIS%%[!0]*}"}"
+NORMALIZED_BUILD_DURATION_MILLIS="${NORMALIZED_BUILD_DURATION_MILLIS:-0}"
+if [ "${#NORMALIZED_BUILD_DURATION_MILLIS}" -gt "${#MAX_BUILD_DURATION_MILLIS}" ] ||
+  { [ "${#NORMALIZED_BUILD_DURATION_MILLIS}" -eq "${#MAX_BUILD_DURATION_MILLIS}" ] &&
+    [ "$NORMALIZED_BUILD_DURATION_MILLIS" -gt "$MAX_BUILD_DURATION_MILLIS" ]; }; then
```
cmur2, line 19: 🔧  These statements are Bash black magic, here it would be good to have a comment at least explaining.

submit-build-status/action.yml
```diff
@@ -77,22 +101,21 @@ runs:
       credentials_json: '${{ inputs.gcp_credentials_json }}'
       token_format: 'access_token'
 
-  - name: Collect resource usage metrics
-    id: monitor
-    shell: bash
-    run: |
-      bash "$GITHUB_ACTION_PATH/stop-and-collect-build-monitor.sh" "$GITHUB_OUTPUT"
```
cmur2, line 84: 🔧  Moving this step will influence existing CPU and memory measurement graphs for existing users. I'd really prefer to keep the compatibility, even if that means that the build duration also includes GCloud login and stuff.

submit-build-status/action.yml
```diff
@@ -69,6 +59,40 @@ runs:
...
+      bash "$GITHUB_ACTION_PATH/stop-and-collect-build-monitor.sh" "$GITHUB_OUTPUT"
+
+  - name: Resolve build duration
+    id: duration
+    shell: bash
+    env:
+      EXPLICIT_BUILD_DURATION_MILLIS: "${{ steps.validate.outputs.build_duration_millis }}"
+      MONITOR_BUILD_DURATION_MILLIS: "${{ steps.monitor.outputs.build_duration_millis }}"
+    run: |
+      BUILD_DURATION_MILLIS=$(bash "$GITHUB_ACTION_PATH/resolve-build-duration.sh" \
```
cmur2, line 79: 🔧 Is a dedicated script needed for 5 lines of Bash code? I'd prefer to keep it inline for short snippets.

submit-build-status/validate-build-duration.sh
cmur2: Also consider inlining the script into the GHA yaml if it is kinda short. Alternatively, move all input validation to a Bash script - for consistency.

start-build-monitor/start-build-monitor.sh
```diff
@@ -2,8 +2,13 @@
 
 INTERVAL=5  # seconds
 # NOTE: these paths are shared with the monitor stop script(s) — if changed, update both
-LOG_FILE=/tmp/_monitor-start.log
-PID_FILE=/tmp/_monitor-start.pid
+LOG_FILE="${BUILD_MONITOR_LOG_FILE:-/tmp/_monitor-start.log}"
```
cmur2, line 5: 💭  I don't really see why these are parameterized? We actually want to hard-code them since they need to stay in sync between the actions, no?

submit-build-status/stop-and-collect-build-monitor.sh
```diff
@@ -8,78 +8,111 @@ fi
...
+
+if [ -f "$START_TIME_FILE" ]; then
+  START_TIME_MILLIS=$(cat "$START_TIME_FILE")
+  END_TIME_MILLIS=$(date +%s%3N)
+  if is_valid_timestamp "$START_TIME_MILLIS" && is_valid_timestamp "$END_TIME_MILLIS"; then
+    BUILD_DURATION_MILLIS=$((10#$END_TIME_MILLIS - 10#$START_TIME_MILLIS))
+    if [ "$BUILD_DURATION_MILLIS" -ge 0 ] && [ "$BUILD_DURATION_MILLIS" -le "$MAX_BUILD_DURATION_MILLIS" ]; then
+      echo "Build duration: ${BUILD_DURATION_MILLIS}ms"
+    else
+      echo "Build duration is outside the supported range (0-${MAX_BUILD_DURATION_MILLIS}ms); duration omitted" >&2
```
cmur2, line 30: ❓ Wait, isn't that validation already done earlier in `validate-build-duration.sh` - this seems duplicated.

submit-build-status/stop-and-collect-build-monitor.sh
```diff
@@ -8,78 +8,111 @@ fi
...
+    echo "Build start or end timestamp is invalid; duration omitted" >&2
+  fi
+else
+  echo "Build start timestamp not found; duration omitted" >&2
+fi
 
 if [ -f "$PID_FILE" ]; then
   MONITOR_PID=$(cat "$PID_FILE")
   kill "$MONITOR_PID" 2>/dev/null || true
-  rm -f "$PID_FILE"
```
cmur2, line 17: 🔧 Why is there such a huge diff that is seemingly unrelated to the purpose of the PR? Can we avoid that?

start-build-monitor/start-build-monitor.sh
```diff
@@ -2,8 +2,13 @@
...
+PID_FILE="${BUILD_MONITOR_PID_FILE:-/tmp/_monitor-start.pid}"
+START_TIME_FILE="${BUILD_MONITOR_START_TIME_FILE:-/tmp/_monitor-start.epoch-millis}"
+
+# Remove state from a previous job before establishing this job's start time.
+rm -f "$PID_FILE" "$LOG_FILE" "$START_TIME_FILE"
+date +%s%3N > "$START_TIME_FILE"
```
cmur2, line 11: 🔧 I raised it before: I don't think we should measure millisecond precision (and thus suggest better accuracy than the measurement actually is) for CI jobs that takes several minutes.

Originally, when I created the BigQuery table I went with higher precision to allow storing Unix timestamps easily in case we need - but for CI we don't really need it yet.

cmur2 (commented):
I still agree with the overall change but I fear this PR touches more places than absolutely necessary to achieve that change. Hence my review got a bit nit-picky, sorry for that and feel free to challenge me: ❓ are (sometimes suggestive) questions 🔧 I would like to see changed 💭 are nice2have

I'd love to see the minimal change we need (docs, code) to get the result we want.

leiicamundi (commented):
.

leiicamundi:
Design note, not a blocker — **this PR should land regardless**, and I am not asking to hold it. But it changes where the duration question ends up, so it is better raised now than after eleven repositories migrate.

@cmur2 asked earlier in this thread:

> did you explore other (implicit) ways of deriving the job/runner start time? From the user's perspective what mostly matters is how long they have to wait for CI feedback.

and the answer was no. There is one, and GitHub already computes it.

## `GET /repos/{owner}/{repo}/actions/runs/{run_id}/jobs`

Returns, per job, with no instrumentation at all: `name` **with the matrix fully expanded**, `status`, `conclusion`, `created_at` (queued), `started_at`, `completed_at`, `runner_name`, `labels`.

Measured against [run 31689628694](https://github.com/camunda/keycloak/actions/runs/31689628694), the same run camunda/keycloak#622 used to validate its wrapper:

| job | API | wrapper submitted |
|---|---|---|
| `build-image (26-hub)` | queued 3s, ran **445s** | 428s |
| `test-postgres-integ (ubuntu-24.04, …25-quay-optimized)` | queued 2s, ran **90s** | 84s |
| `test-postgres-integ (aws-core-2-default, …)` | **queued 642s**, ran 326s | ~326s at best |

The 4–7% under-report on execution time is the small part, and it is the part this PR fixes.

The large part is the third row. **642 seconds of queue time — 66% of what the developer actually waits — is structurally invisible to any in-job measurement**, whichever anchor is chosen. The job does not exist yet. No pid file, no `$GITHUB_ACTION_PATH`, and no future refinement of either can reach it.

That matters specifically because of what the data is for. camunda/team-infrastructure#1051 is runner right-sizing, and 642s of queue on `aws-core-2-default` is an under-capacity signal that execution time alone does not reveal — a pool can look perfectly sized on duration while developers wait ten minutes for a slot.

## What else comes free

- `job_name` already carries the expanded matrix, which removes the need for `job_name_override` and the separator convention it forces on callers
- `runner_name` and `labels`, currently read from `/home/runner/.camunda-arc-runner-info/runs-on` — a Camunda-specific file that only exists on ARC runners
- it works retroactively, on jobs nobody onboarded, and on jobs that fail before reaching their last step

## The split this suggests

- **duration, status, queue time, runner** → a `workflow_run`-triggered collector reading the API. `workflow-run-stats` in this repository already does API-based collection over a window, so the pattern is established here.
- **CPU / memory / network** → genuinely requires in-job sampling. `start-build-monitor` + `submit-build-status` stay, opt-in, on the jobs where right-sizing actually needs them.

Costs, so this is not one-sided: `workflow_run` workflows run from the default branch and need `actions: read`; `run_attempt` needs handling for reruns; it adds one job per run; and it is a one-time discontinuity in `build_status_v2` — durations grow and a queue column appears. That last one is @cmur2's continuity concern again, but as a single well-defined migration rather than per-repository drift.

## Why I am not asking to hold this PR

The resource-metric path needs `start-build-monitor` either way, and this PR is a clear improvement over eleven wrappers each computing their own duration — including the stale-state clearing, which is a real bug fix. It should go in.

What I would suggest is not treating the in-job duration as the final answer while the eleven repositories migrate, since each migration is a chance to point them somewhere better. camunda/keycloak#622 and camunda/camunda-deployment-references#2975 are keeping their vendored copies for now with a `TODO` marking them as temporary, which keeps that door open.

Happy to open a separate issue against #1051 with the collector sketch if that is more useful than a comment here.

leiicamundi:
Filed the collector sketch as camunda/team-infrastructure#1128 so it does not sit on this PR.

Two things I found while writing it up that strengthen the case, both from the tracking issues rather than from me:

- team-infrastructure#1080 measures the coverage gap: **only 4 of 30 repositories** consuming self-hosted minutes appear in `build_status_v2`, leaving 2.42M SHR minutes — 10.5% of the total — uncovered. Closing that by instrumentation scales with the number of jobs; via the API it scales with the number of repositories.
- the same issue asks for a durable workflow identity because the billing report keys on `workflow_path` while `build_status_v2` stores `workflow_name`. `GET /repos/{owner}/{repo}/actions/runs/{run_id}` already returns `path` — the billing key itself — plus a numeric `workflow_id` that survives renames. This action currently derives `workflow_ref` by string-manipulating `$GITHUB_WORKFLOW_REF` (`submit-build-status/action.yml:105-106`).

None of this changes my position on this PR: it should land. The utilisation metrics are the part that genuinely requires in-job sampling, and this PR is a clear improvement over eleven wrappers each computing their own duration.

cmur2:
Hi, just FYI I closed https://github.com/camunda/team-infrastructure/issues/1128 since without coordination suggesting a whole rearchitecture of the CI Analytics system to get more accurate build duration measures is disproportionate.

My original question "did you explore other (implicit) ways of deriving the job/runner start time?" was bounded by the current (implicit) architecture context (and within that using different signs/triggers/files), so please do not pull it out of context and inflate it to a bigger question that it was.

leiicamundi:
@cmur2 your call, and I am not reopening the substance.

On "without coordination" I would push back: the issue *was* the coordination. Asynchronous, it did not claim your attention while you were mid-review here, and it cost you a minute to decline. Raising it only in this PR thread would have buried an architecture question inside an unrelated code review — which is what you are objecting to.

If the objection is the channel, I disagree. If it is the sequencing, that I should have asked before investing the effort and formalising it against your roadmap, that is fair and I take it.

On the quote the framing was mine and it was wrong: your question was bounded to signals inside the current design, and I used it as the opening for something wider. But "more accurate build duration measures" was not the driver — it was the eleven copies and what they cost to keep alive. Accuracy was a side effect I over-sold.

Separately, so they do not go out with the issue: `test_build_duration.sh` is not wired into pre-commit or any workflow, and the second-precision change agreed on 21 July is not in the branch.

maxdanilov:
@Kerruba is this PR still relevant?

Issue #778, opened 2026-08-10
actionlint action: installer is curl|bash from a mutable ref, with no checksum


leiicamundi:
## Problem

`actionlint/action.yml` installs actionlint like this:

```bash
bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash) $ACTIONLINT_VERSION
```

Three things are unverified in that one line:

1. **The script comes from a mutable ref.** `.../rhysd/actionlint/main/scripts/...` is
   whatever is on upstream's default branch at the moment the job runs. It is not pinned
   to a tag or a commit.
2. **It is piped straight into `bash`.** Whatever that URL returns executes with the job's
   privileges.
3. **`curl` has no failure or transport flags** — no `-f`, no `--proto '=https'`, no
   `--tlsv1.2`. On an HTTP error the body is still handed to `bash`.

And the script it runs does not verify what it downloads either:

```bash
curl -L "${url}" | tar xvz -C "$target_dir" actionlint
```

No checksum, no signature. So the binary that ends up linting workflows is whatever the
release URL served.

## Why this matters more than usual here

This is a shared action. A change to that upstream script — or anything able to influence
what that URL returns — reaches every repository and every workflow that calls it, at
once.

In `camunda/infraex-common-config` specifically, the step runs in the same job that
produces the auto-fix patch which a later job commits back to the pull request with a
GitHub App token (camunda/infraex-common-config#550). Code executing in that job can
influence what gets committed.

## Suggested fixes, cheapest first

**1. Pin the installer script.** Replace `main` with a tag or commit SHA:

```bash
bash <(curl -fsSL --proto '=https' --tlsv1.2 \
  "https://raw.githubusercontent.com/rhysd/actionlint/v${ACTIONLINT_VERSION}/scripts/download-actionlint.bash") \
  "$ACTIONLINT_VERSION"
```

Using the same ref as the version being installed keeps the two consistent and removes
the mutable-branch dependency. `-f` also stops an error page from being executed.

**2. Skip the installer entirely.** The release assets can be fetched directly and
checked against the `SHA256SUMS` published on the same release, or the official
`docker.io/rhysd/actionlint:<version>` image can be used, which is content-addressable by
digest.

## Smaller items in the same file

- `- uses: actions/checkout@v7` is a mutable tag. Every other `uses:` in the repositories
  consuming this action is SHA-pinned; this one is not.
[...]

PR #779, opened 2026-08-11
fix: pin vault-action and create-github-app-token to full-length SHAs

leiicamundi:
## Problem

`generate-github-app-token-from-vault-secrets/action.yml` references three actions by tag:

* `hashicorp/vault-action@v4.0.0` (twice)
* `actions/create-github-app-token@v3`

Any repository that enables **Settings → Actions → General → Require actions to be pinned to a full-length commit SHA** can no longer run *a single job* that resolves this action. GitHub evaluates the policy over the whole dependency tree at *Set up job*, before any step executes:

```
##[error]The actions hashicorp/vault-action@v4.0.0 and actions/create-github-app-token@v3
are not allowed in camunda/infraex-terraform because all actions must be pinned to a
full-length commit SHA.
```

A consumer pinning its own call sites does not help: the offending reference lives here. In `camunda/infraex-terraform` this currently takes out `lint`, `terraform-drift-detection` and `certificate-renewal`, the last one being a scheduled Let's Encrypt renewal.

## Change

Pin the three references to the commit the tag points at today, with the version kept in a trailing comment so Renovate keeps updating them:

| Reference | SHA |
| --- | --- |
| `hashicorp/vault-action` v4.0.0 | `892a26828f195e65540a40b4768ae4571f51ebfc` |
| `actions/create-github-app-token` v3.2.0 (what `v3` resolves to) | `bcd2ba49218906704ab6c1aa796996da409d3eb1` |

No behaviour change: same code, immutable reference.

## Scope

Deliberately limited to the action that blocks consumers today. The rest of the repository is tag-pinned as well and will hit the same wall as the policy spreads, but that is a much larger change and belongs in its own PR. Related: #778, which covers the `actionlint` action (`curl | bash` from a mutable ref, plus `actions/checkout@v7`).

### DIFF
--- a/generate-github-app-token-from-vault-secrets/action.yml
+++ b/generate-github-app-token-from-vault-secrets/action.yml
@@ -125,7 +125,7 @@ runs:
 
   - name: Import the GitHub App ID from vault
     id: app-id
-    uses: hashicorp/vault-action@v4.0.0
+    uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
     with:
       url: ${{ inputs.vault-url }}
       method: ${{ inputs.vault-auth-method }}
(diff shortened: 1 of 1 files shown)


kellervater (approved):
lgtm under the assumption you double checked the SHAs. copilot/claude sometimes hallucinates them.

leiicamundi:
Thanks @kellervater - I have performed a double check on the sha, they do match.

PR #781, opened 2026-08-11
ci: pin every GitHub Action to a commit SHA and enforce it with zizmor

kellervater:
## Summary

Pins **every** action used in this repo — workflows and composite actions, **including GitHub's own `actions/*`** — to a 40-character commit SHA, with the full version in a trailing comment. Enforced by `zizmor` as a pre-commit hook, running locally and in the existing `pre-commit` CI workflow.

Two reasons a tag is not good enough here: a tag is a mutable pointer a compromised upstream can force-push, and several consuming repositories already enforce GitHub's *Require actions to be pinned to a full-length commit SHA* setting, which exempts no namespace and is evaluated over the whole dependency tree. An action left on a tag here refuses the **consumer's** job at `Set up job`, before any step runs.

| | |
|:--|:--|
| **Policy** | `hash-pin` everything, including `actions/*` and `github/*`. Only `camunda/*` is exempt (`@main` self-reference contract). |
| **Scale** | 75 occurrences pinned; 37 unique refs; 34 files |
| **Verified** | 38/38 pins re-resolved against upstream, **0 mismatches**; 35/35 replaced refs still point at the SHA they were pinned to; gate confirmed to fail on an unpinned ref in both workflows and composite actions |
| **Behavior change** | None. Every SHA is what its ref resolved to already, and the diff touches `uses:` lines only. |
| **Stricter than** | `camunda/camunda` and `infra-core`, deliberately — see below |
| **⚠️ Known gap** | Three `@main` self-references remain refused downstream — see below |

<details>
<summary><b>Why <code>actions/*</code> is not exempt here, unlike in <code>camunda/camunda</code> and <code>infra-core</code></b></summary>

Those repos exempt `actions/*` and `github/*` as sitting inside the same trust boundary as the runners. Sound for them; it does not carry over here, because the binding constraint is not our own threat model but GitHub's setting, which grants no namespace exemption:

> all actions must be pinned to a full-length commit SHA to be used. This includes actions from your organization and actions authored by GitHub. Reusable workflows can still be referenced by tag.
>
> — [Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository)

Evaluated over the **whole dependency tree**, refused at `Set up job`. A consumer cannot work around it by pinning a different revision of this repo. A `ref-pin` policy for `actions/*` would be weaker than the policy already binding on the repositories this gate exists to protect: green here, broken downstream.

This repo is the wrong place for that gap — it is public, its composite actions run inside other teams' workflows, and consumers reference it at `@main`.

</details>

<details>
<summary><b>⚠️ Known gap: <code>camunda/*</code> <code>@main</code> self-references</b></summary>

Note the docs carve-out is for **reusable workflows**, not organization ownership — so `camunda/*` gets no exemption on "it's ours" grounds either. Every `camunda/*` reference here is a step-level composite action.

Three composite actions self-reference this repo at `@main`:

| File | Line |
|---|---|
| `setup-yarn-cache/action.yml` → `is-cache-enabled` | 43 |
| `fossa/info/action.yml` → `setup-gh-cli` | 49 |
| `teams/infra/configure-maintenance-pull-request/action.yml` → `configure-pull-request` | 97 |

**These three are still refused for a consumer enforcing the setting, even after this PR.** So this PR unblocks consumers reaching `build-docker-image`, `rerun-failed-run` and `common-tooling`, but not those three.

Closing the gap means giving up the `@main` contract — every internal change becomes a merge-then-bump two-step — which is a repository-design decision, not something to settle in a lint policy. Documented as an open gap in `.github/zizmor.yml` and the docs rather than silently ignored.

</details>

<details>
<summary><b>Changes</b></summary>

1. `.github/zizmor.yml` — `unpinned-uses` policy; the six other audits that fire here are explicitly disabled with backlog counts recorded as standing TODOs (155 findings across the six, largest 125 `template-injection`)
2. `zizmor` pre-commit hook in `.pre-commit-config.yaml` — picked up by the existing `pre-commit` workflow, so no CI change needed
3. 75 references pinned (46 third-party, then 29 `actions/*`)
4. `docs/github-actions-pinning.md` — policy, rationale, rejected alternatives, known gap
5. README section on pinning
6. `.github/renovate.json5` — comment documenting the full-version-comment coupling (no behavior change)

</details>

<details>
<summary><b>Counts</b></summary>

| Measure | Count |
|:--------|------:|
| Occurrences newly pinned by this PR | **75** (46 third-party + 29 `actions/*`) |
| Pre-existing pins on `main` | 4 |
| Total pinned occurrences on the branch | **79** |
| Unique pinned refs | **37** |
| YAML files containing at least one pin | 31 |
| Files changed | 34 |

</details>

<details>
<summary><b>✅ Why this adds no Renovate toil</b></summary>

Pin comments record the **full** version (`# v4.6.0`), never a floating major (`# v4`). Load-bearing, because Renovate derives the update type from the comment:

| Comment | Upstream ships v4.6.0 | Update type | Auto-merged? |
|:--------|:----------------------|:------------|:-------------|
| `# v4.5.0` | comment and SHA both updated | `patch` | ✅ yes |
| `# v4` | `v4` tag moves, comment unchanged | `digest` | ❌ no, costs a review |

`.github/renovate.json5` auto-merges `minor`/`patch` for the `github-actions` manager, not `digest`. Floating-major comments would have turned every routine upstream release into a hand-reviewed PR for zero security gain, and destroyed the signal value of a digest-only PR — which now genuinely means an upstream tag moved without a release.

</details>

<details>
<summary><b>Verification</b></summary>

Two independent passes, both reading pins back out of the working tree rather than from the table used to write them:

- **Do the SHAs match their comments?** 38/38 unique pins re-resolved via `git ls-remote`, annotated tags dereferenced: **38 exact, 0 mismatches.**
- **Did anything silently change version?** Stronger check, and the one that matters: each replaced ref re-resolved to confirm it still points at the SHA it was pinned to. **35/35 identical.** Matching a tag only proves the SHA is real; this proves the pin is the same code that ran before. `actions/create-github-app-token@v3` resolves to `bcd2ba49…`, the same SHA already hand-pinned in `generate-github-app-token-from-vault-secrets`. (`8398a7/action-slack@v3` is a *branch*, not a tag; its head is the pinned SHA, which also carries `v3.19.0`.)
- **Nothing else moved.** The diff across `.github/workflows/*` and every `action.yml` contains `uses:` lines exclusively — no functional change hidden in 34 files.
- **Gate bites.** Reverting `actions/create-github-app-token`, `actions/setup-node` or `mikefarah/yq` to a tag each fails the hook; `camunda/*@main` still passes, as intended. Works in workflows **and** composite `action.yml` files.
- **Backlog counts re-derived** from `zizmor --format json`: 155 across the six disabled audits (as recorded), 238 for the full default audit set on `main`.
- `pre-commit run -a` green on actionlint, zizmor, renovate-config-validator.

</details>

<details>
<summary><b>Out of scope, tracked separately</b></summary>

- **`pre-commit/action` resolves `actions/cache@v4` internally.** If the setting is ever enabled on *this* repo, the workflow enforcing the policy would itself be refused at `Set up job`. Needs the inlined-`pre-commit` treatment used in `infraex-common-config` (#557, #561). Own issue.
- **`8398a7/action-slack` is an archived upstream** (2 uses). Pinned, but it will never receive security fixes. Replacing it is a behavioral change.
- **`codex-/return-dispatch` v4** left to #772; this PR pins it at v3.0.3, i.e. what runs today.

</details>

<details>
<summary><b>Why no ADR</b></summary>

This repo has no `adr/` convention, and it is public while `infra-core` is private — so a pointer to that ADR would be unreadable to exactly the external consumers who most need the rationale. The decision record, including rejected alternatives, is folded into `docs/github-actions-pinning.md`, which links only to public sources.

</details>

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,18 +12,18 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 5
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Setup Python
-        uses: actions/setup-python@v7
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: "3.x"
 
-      - uses: asdf-vm/actions/install@v4
+      - uses: asdf-vm/actions/install@b7bcd026f18772e44fe1026d729e1611cc435d47 # v4.0.1
 
       - run: |
           echo "::add-matcher::actionlint/actionlint-matcher.json"
 
-      - uses: pre-commit/action@v3.0.1
+      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
         with:
           extra_args: --all-files --verbose
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,91 @@
+---
+# =============================================================================
+# zizmor.yml — GitHub Actions supply-chain policy for infra-global-github-actions
+#
+# Consumed by the `zizmor` pre-commit hook (see .pre-commit-config.yaml), which
+# runs locally on `pre-commit run -a` and in CI via the pre-commit workflow.
+#
+# This repository is public and its composite actions are consumed by other
+# Camunda teams, so an unpinned third-party action here is not just our own
+# exposure — every downstream repository inherits it transitively.
+#
+# Scope: this repository currently adopts a single zizmor audit, `unpinned-uses`.
+# Every other audit that fires here is explicitly disabled below. The audits are
+# disabled rather than left on because this repo has pre-existing findings for
+# all of them; a lint gate that is red on arrival gates nothing. Each
+# `disable: true` entry is a standing invitation to fix the underlying findings
+# and drop the entry.
+#
+# Docs: https://docs.zizmor.sh/configuration/
+# =============================================================================
+
+rules:
+  unpinned-uses:
+    config:
+      # `policies` maps action-owner globs to a pinning level. Globs are matched
+      # most-specific-first, so the `*` catch-all only applies to owners that no
+      # other entry claims.
+      #
+      #   any       — no constraint; any ref (SHA, tag, branch) is accepted.
+      #   ref-pin   — must carry an explicit ref, symbolic (tag/branch) or SHA.
+      #   hash-pin  — must be a 40-character commit SHA. Tags and branches fail.
+      #
+      # Local references (`uses: ./common-tooling`) are same-repo and are skipped
+      # by this audit entirely, so they need no policy entry.
+      policies:
+        # Everything is hash-pinned, including GitHub's own actions.
+        #
+        # camunda/camunda and infra-core exempt `actions/*` and `github/*` as
+        # "same trust boundary as the runners". That reasoning does not carry
+        # over to this repository, because the binding constraint here is not
+        # our own threat model but GitHub's `Require actions to be pinned to a
+        # full-length commit SHA` setting, which several consumers already
+        # enforce. That setting grants no namespace exemptions:
+        #
+        #   "all actions must be pinned to a full-length commit SHA to be used.
+        #    This includes actions from your organization and actions authored
+        #    by GitHub. Reusable workflows can still be referenced by tag."
+        #   -- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository
+        #
+        # It is evaluated over the whole dependency tree, so an `actions/foo@v1`
+        # left on a tag inside a composite action here refuses the consumer's
+        # job at `Set up job`, before any step runs, and the consumer cannot
+        # work around it by choosing a different revision of this repository.
+        # A `ref-pin` policy for `actions/*` would therefore be weaker than the
+        # policy already binding on the repositories this gate exists to
+        # protect -- green here, broken downstream.
+        "*": hash-pin
+
+        # The one exemption, and it is a deliberate trade-off rather than a
+        # trust judgement. This repository references itself at `@main`
+        # (setup-yarn-cache -> is-cache-enabled, fossa/info -> setup-gh-cli,
+        # configure-maintenance-pull-request -> configure-pull-request), which
+        # is the contract its consumers rely on: they get the current `main`
+        # without a bump PR. Hash-pinning those would make every internal
+        # change a two-step merge-then-bump.
+        #
+        # KNOWN GAP: because the GitHub setting walks the whole tree, those
+        # three `@main` self-references are still refused for a consumer that
+        # enforces it, even after this change. Closing that means giving up the
+        # `@main` contract, which is a repository-design decision and is
+        # tracked separately rather than settled here.
+        camunda/*: ref-pin
+
+  # --- audits not adopted (yet) ----------------------------------------------
+  # Findings below are pre-existing and unrelated to tag mutability. Enabling
+  # any of them means fixing its findings in the same change. The counts are
+  # from the adopting change (`zizmor 1.29.0 --offline --no-config`, 155
+  # findings across these six audits) and are recorded to show the size of each
+  # backlog, not to be kept current.
+  archived-uses:            # 2 findings — both 8398a7/action-slack, upstream is archived
+    disable: true
+  artipacked:               # 8 findings
+    disable: true
+  excessive-permissions:    # 8 findings
+    disable: true
+  github-app:               # 6 findings
+    disable: true
+  github-env:               # 6 findings
+    disable: true
+  template-injection:       # 125 findings
+    disable: true
--- a/common-tooling/action.yml
+++ b/common-tooling/action.yml
@@ -211,14 +211,14 @@ runs:
   - name: Set up QEMU
     if: ${{ inputs.qemu-enabled == 'true' }}
     id: qemu
-    uses: docker/setup-qemu-action@v4
+    uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
     with:
       image: ${{ inputs.qemu-image }}
       platforms: ${{ inputs.qemu-platforms }}
   # Docker Buildx
   - name: Set up Docker Buildx
     if: ${{ inputs.buildx-enabled == 'true' && ( env.buildx_missing == 'true' || inputs.overwrite == 'true' ) }}
-    uses: docker/setup-buildx-action@v4
+    uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
     with:
       version: ${{ inputs.buildx-version }}
       driver: ${{ inputs.buildx-driver }}
@@ -243,7 +243,7 @@ runs:
       export PATH="$PATH:`yarn global bin`"
   # Install node
   - name: Install NodeJS
-    uses: actions/setup-node@v7
+    uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
     if: ${{ inputs.node-enabled == 'true' && ( env.node_missing == 'true' || inputs.overwrite == 'true' ) }}
     with:
       always-auth: ${{ inputs.node-always-auth }}
@@ -260,7 +260,7 @@ runs:
   # Setup Maven
   - name: Setup Maven Action
     if: ${{ inputs.java-enabled == 'true' && ( env.java_missing == 'true' || inputs.overwrite == 'true' ) }}
-    uses: s4u/setup-maven-action@v1.20.0
+    uses: s4u/setup-maven-action@ba34de01b7f4ba2ab8e2860df8993a29f4477056 # v1.20.0
     with:
       checkout-enabled: false
       java-version: ${{ inputs.java-version }}
@@ -272,7 +272,7 @@ runs:
 
   # Install Python, it's absent on self-hosted runners
   - name: Install Python3 (std install)
-    uses: actions/setup-python@v7
+    uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
     # actions/setup-python doesn't yet support ARM
     if: ${{ !contains(runner.arch, 'arm') && inputs.python-enabled == 'true' && ( env.python_missing == 'true' || inputs.overwrite == 'true' ) }}
     with:
@@ -287,7 +287,7 @@ runs:
       allow-prereleases: ${{ inputs.python-allow-prereleases }}
 
   - name: Install Python3 (arm install)
-    uses: deadsnakes/action@v3.2.0
+    uses: deadsnakes/action@e640ac8743173a67cca4d7d77cd837e514bf98e8 # v3.2.0
     # actions/setup-python doesn't yet support ARM, see https://github.com/actions/setup-python/issues/705#issuecomment-1756948951
     if: ${{ contains(runner.arch, 'arm') && inputs.python-enabled == 'true' && ( env.python_missing == 'true' || inputs.overwrite == 'true' ) }}
     with:
--- a/rerun-failed-run/action.yml
+++ b/rerun-failed-run/action.yml
@@ -108,7 +108,7 @@ runs:
 
     - name: Import Secrets
       id: vault-secrets
-      uses: hashicorp/vault-action@v4.0.0
+      uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
       with:
         url: ${{ inputs.vault-addr }}
         method: ${{ inputs.vault-auth-method }}
@@ -123,7 +123,7 @@ runs:
 
     - name: Generate a GitHub token for infra-rerun camunda/infra-global-github-actions
       id: github-token
-      uses: actions/create-github-app-token@v3
+      uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
       with:
         app-id: ${{ steps.vault-secrets.outputs.RETRIGGER_APP_ID }}
         private-key: ${{ steps.vault-secrets.outputs.RETRIGGER_APP_KEY }}
@@ -153,7 +153,7 @@ runs:
       shell: bash
 
     - name: Trigger retry-worfklow-run.yml
-      uses: codex-/return-dispatch@v3
+      uses: codex-/return-dispatch@08365b69369e9eb639ce4c4cbbd29598bc52ca3a # v3.0.3
       id: dispatch
       with:
         owner: camunda
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -53,6 +53,14 @@
       matchPackageNames: ["renovatebot/pre-commit-hooks"],
       schedule: ["on Saturday"],
     },
+    // Third-party actions are SHA-pinned (see docs/github-actions-pinning.md).
+    // Renovate derives the update type from the version comment on the `uses:`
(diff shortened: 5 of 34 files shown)


.github/zizmor.yml
```diff
@@ -0,0 +1,69 @@
...
+      #   any       — no constraint; any ref (SHA, tag, branch) is accepted.
+      #   ref-pin   — must carry an explicit ref, symbolic (tag/branch) or SHA.
+      #   hash-pin  — must be a 40-character commit SHA. Tags and branches fail.
+      #
+      # Local references (`uses: ./common-tooling`) are same-repo and are skipped
+      # by this audit entirely, so they need no policy entry.
+      policies:
+        # GitHub-owned actions, treated as part of the platform we already trust
+        # to run our jobs. Same trust boundary as camunda/camunda and infra-core.
+        actions/*: ref-pin
```
leiicamundi: 🔴 `actions/*: ref-pin` is what keeps this PR from delivering its stated goal.

GitHub's **Require actions to be pinned to a full-length commit SHA** does not exempt `actions/*`. A repository with that setting on refuses, verbatim:

```
The actions hashicorp/vault-action@v4.0.0 and actions/create-github-app-token@v3
are not allowed in <repo> because all actions must be pinned to a full-length
commit SHA.
```

`actions/create-github-app-token@v3` is named explicitly. The refusal happens at `Set up job` before any step runs, and it is evaluated over the **whole** dependency tree — so a consumer cannot work around it by picking another revision of this repo.

The gate installed here is therefore weaker than the policy it is meant to satisfy. Add `actions/foo@v1` to a composite action, zizmor stays green, `main` moves, and consumers start failing at `Set up job` with no Renovate PR anywhere in the chain. That is the propagation path the description itself calls out.

Verified on this branch: `mikefarah/yq` back to a tag → hook fails as intended. Comment added next to `actions/create-github-app-token@v3` → zizmor still green.

Fix: `actions/*: hash-pin`, `github/*: hash-pin`. `camunda/*` too, unless the `@main` self-reference contract forbids it — if so, worth saying here, since that contract is what makes the blast radius immediate.

If the scope is deliberate, narrow the description instead. It reads as if pinning here is sufficient downstream, and for `actions/*` it is not.

rerun-failed-run/action.yml
```diff
@@ -108,7 +108,7 @@ runs:
 
     - name: Import Secrets
       id: vault-secrets
-      uses: hashicorp/vault-action@v4.0.0
+      uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
```
leiicamundi, line 111: 🔴 L126 `actions/create-github-app-token@v3` stays unpinned under `actions/*: ref-pin`, so this action is still refused wherever the SHA-pinning setting is on.

`build-docker-image` comes out fully clean, this one does not. The PR therefore fixes some composite actions and not others, which is worth stating explicitly rather than leaving consumers to find out at `Set up job`.

common-tooling/action.yml
```diff
@@ -211,14 +211,14 @@ runs:
...
-    uses: docker/setup-qemu-action@v4
+    uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
     with:
       image: ${{ inputs.qemu-image }}
       platforms: ${{ inputs.qemu-platforms }}
   # Docker Buildx
   - name: Set up Docker Buildx
     if: ${{ inputs.buildx-enabled == 'true' && ( env.buildx_missing == 'true' || inputs.overwrite == 'true' ) }}
-    uses: docker/setup-buildx-action@v4
+    uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
```
leiicamundi, line 221: 🔴 Same gap as `rerun-failed-run`: L246 `actions/setup-node@v7` and L275 `actions/setup-python@v7` stay unpinned, so `common-tooling` is still refused wherever the SHA-pinning setting is on.

.github/workflows/pre-commit.yml
```diff
@@ -19,11 +19,11 @@ jobs:
...
           python-version: "3.x"
 
-      - uses: asdf-vm/actions/install@v4
+      - uses: asdf-vm/actions/install@b7bcd026f18772e44fe1026d729e1611cc435d47 # v4.0.1
 
       - run: |
           echo "::add-matcher::actionlint/actionlint-matcher.json"
 
-      - uses: pre-commit/action@v3.0.1
+      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
```
leiicamundi, line 27: 🟡 `pre-commit/action` is pinned, but resolves `actions/cache@v4` internally, and the setting is evaluated over the whole tree. Enable it on this repo and the workflow enforcing the policy is itself refused at `Set up job`.

Known-good remedy: install `pre-commit` via asdf (which also stops `pip install pre-commit` overriding the `.tool-versions` pin), declare `actions/cache` explicitly to keep the caching the action provided, then `run: pre-commit run --all-files`. Out of scope here — worth an issue so the gate cannot fail before it runs.

leiicamundi (commented):
Verified independently, not a read-through.

**28/28 pins re-resolved from the working tree via `git ls-remote`, tags dereferenced. 0 mismatches.** Matches your count. Core of the PR is clean.

Gate bites where you say it does: `mikefarah/yq` back to a tag → hook fails. zizmor does scan nested composites — `rerun-failed-run/check-annotations-and-logs/action.yml` shows in its output.

## Blocker: `actions/*: ref-pin`

GitHub's **Require actions to be pinned to a full-length commit SHA** does not exempt `actions/*`. A repository with that setting on refuses `actions/create-github-app-token@v3` by name, at `Set up job`, before any step runs — and the check is evaluated over the **whole** dependency tree, so a consumer cannot escape it by picking another revision of this repo.

That makes the gate weaker than the policy it is meant to satisfy. Concretely, after this PR:

| composite action | state |
|---|---|
| `build-docker-image` | fully pinned, clean |
| `rerun-failed-run` | L126 `actions/create-github-app-token@v3` |
| `common-tooling` | L246, L275 `actions/setup-node@v7`, `setup-python@v7` |
| `preview-env/*`, `setup-yarn-cache`, `add-bug-to-quality-board`, `configure-pull-request`, `actionlint`, `assert-camunda-git-emails`, `check-copilot-commits` | `actions/*` on tags |

So the change lands unevenly: one composite action becomes usable under that setting, the rest stay refused, and nothing in the repo signals the difference. A future `actions/foo@v1` added to any composite here passes the hook and breaks consumers silently.

Two ways out, both fine by me:

1. `actions/*: hash-pin` + `github/*: hash-pin`. ~15 more pins, gate matches what is actually enforced, uniform result.
2. Keep the scope, narrow the description — it currently reads as though pinning here is sufficient downstream.

## Minor

- `pre-commit.yml` L27: `pre-commit/action` resolves `actions/cache@v4`. Enable the setting on this repo and the workflow enforcing the policy is itself refused at `Set up job`. Worth an issue, not this PR.
- 🔵 Counts don't reconcile: description says *46 refs / 22 files*, verification says *28/28*. Actual on branch: **48** pin lines added, **50** occurrences, **28** unique, **27** files. Substance right, headline numbers off.
- Agree on the rest: `8398a7/action-slack` and `codex-/return-dispatch` v4 out of scope, six audits disabled with backlog counts beats a red gate, no ADR since the repo is public.

## Verdict

Approvable on correctness. Pins right, gate works, nothing silently up- or downgraded. Holding approval only on the `actions/*` scope, since it decides whether this PR is one change or two.

kellervater:
hey @leiicamundi thanks for your review.
this PR though wasn't yet ready for review. I just haven't had time to verify what Claude produced here yet.
I will take your review into consideration and I will also be more cautious on opening drafts rather than PRs before filing them. 
Thanks for your input.

kellervater:
> ⚠️ **This is an automated reply written by Claude**, acting on @kellervater's instruction. Patrick will review this work himself once it is finished and he has capacity, and will file the PR properly at that point. **Please hold off on further reviews until he officially files it** — this was opened as a non-draft by mistake and was never ready for review. Apologies for the wasted round.

Thanks for the review — the central finding was correct and has been fixed in b429322.

## `actions/*: ref-pin` — fixed, you were right

Verified the claim independently against GitHub's docs rather than taking it on the error message:

> all actions must be pinned to a full-length commit SHA to be used. This includes actions from your organization and actions authored by GitHub. Reusable workflows can still be referenced by tag.
>
> — [Managing GitHub Actions settings for a repository](``https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository)``

Took **option 1**: `actions/*` and `github/*` now fall under the `hash-pin` catch-all. 29 further occurrences pinned across 9 unique refs. Every SHA is what its tag resolves to today, so nothing changes behaviourally — `actions/create-github-app-token@v3` resolves to `bcd2ba49…`, the same SHA already hand-pinned in `generate-github-app-token-from-vault-secrets`.

Your exact regression case now fails as it should:

```
$ # actions/create-github-app-token reverted to @v3
$ zizmor --offline --config=.github/zizmor.yml .
error[unpinned-uses]: unpinned action reference
38 findings (37 suppressed): 1 high
exit=14
```

`rerun-failed-run` L126 and `common-tooling` L246/L275 are pinned, so those two threads are resolved.

## One correction on `camunda/*`, and a gap worth your attention

You suggested `camunda/*: hash-pin` too. It stays on `ref-pin`, because of the `@main` self-reference contract — but your instinct that this is the load-bearing part was right, and the reason is worse than "it's ours". The docs carve-out is for **reusable workflows**, not for organization ownership, and every `camunda/*` reference here is a step-level composite action. So organization ownership buys no exemption at all.

Concretely, three composite actions self-reference at `@main`:

| File | Line |
|---|---|
| `setup-yarn-cache/action.yml` → `is-cache-enabled` | 43 |
| `fossa/info/action.yml` → `setup-gh-cli` | 49 |
| `teams/infra/configure-maintenance-pull-request/action.yml` → `configure-pull-request` | 97 |

Since the setting walks the whole tree, **those three are still refused for a consumer enforcing it, even after this PR** — `setup-yarn-cache` most notably. So for your sequencing: this PR unblocks consumers that reach `build-docker-image`, `rerun-failed-run` and `common-tooling`, but *not* ones that reach those three. Worth checking your four repos against that list before you re-pin. Closing that gap means giving up the `@main` contract, which is Patrick's call and is documented as an open gap in `.github/zizmor.yml` rather than silently ignored.

## Counts — both of us were right

They measure different things, and the delta reconciles exactly: your 48 lines / 24 files counted the two example `uses:` lines I added to `README.md` and `docs/github-actions-pinning.md`; my 46 / 22 counted only workflow and composite-action changes. Description now states one unambiguous table. Current state: **75 occurrences pinned by this PR** (46 third-party + 29 `actions/*`), 79 total on the branch, **37 unique refs, 37/37 re-resolved against upstream, 0 mismatches**.

## `pre-commit/action` → `actions/cache@v4`

Agreed, and agreed it is out of scope here. Recorded in the description as a follow-up; Patrick will decide whether to file it against this repo with the `infraex-common-config` #557/#561 treatment.

---
_Generated by [Claude Code](https://claude.ai/code)_

kellervater:
> ⚠️ **Automated reply written by Claude**, per @kellervater's instruction. He will review this work himself once it is finished and has capacity, and will file the PR properly then. **Please hold off on further reviews until he officially files it.**

Heads-up before you sequence anything on #782: **both halves are already on this branch**, as of `00e703e`. I think you branched off `main` just as that landed, so the "no shared lines" premise no longer holds — the two PRs now touch the same lines in the same files and will conflict.

Current head of this branch:

```
$ git show 00e703e:.github/zizmor.yml | grep -E '"\*"|camunda/\*'
        "*": hash-pin
        camunda/*: ref-pin

$ # any actions/* or github/* still on a tag?
$ git grep -nE "uses:\s*'?(actions|github)/" 00e703e -- '*.yml' '*.yaml' | grep -vE '@[0-9a-f]{40}'
(no output)
```

So `actions/*` and `github/*` are already under the `hash-pin` catch-all, and there is no unpinned `actions/*` left to reopen the gap. Your closing concern — "nothing stops the next `actions/foo@v1`" — is what that policy change already covers, and the regression test for it is in the description.

## Your pins independently confirm mine

Worth stating plainly, because it is the strongest evidence in this whole thread: your 10 unique refs and their version comments are **byte-identical** to what landed here, resolved separately by each of us.

| ref | SHA | comment |
|---|---|---|
| `actions/checkout` | `3d3c42e5…` | `# v7.0.1` |
| `actions/setup-python` | `5fda3b95…` | `# v7.0.0` |
| `actions/setup-node` | `82076278…` | `# v7.0.0` |
| `actions/github-script` | `3a2844b7…` | `# v9.0.0` |
| `actions/create-github-app-token` | `bcd2ba49…` | `# v3.2.0` |
| `actions/add-to-project` | `5afcf98f…` | `# v2.0.0` |
| `actions/cache` + `cache/restore` | `55cc8345…` | `# v6.1.0` |
| `actions/upload-artifact` | `043fb46d…` | `# v7.0.1` |
| `actions/download-artifact` | `3e5f45b2…` | `# v8.0.1` |

Two independent resolutions, 10/10 agreement on both SHA and full-semver comment, and we independently arrived at the same "29 references, 18 files" for the `actions/*` half. Your `create-github-app-token@bcd2ba49 # v3.2.0` control against #779 holds here too.

## The call is Patrick's

Since #781 already carries the pins *and* the policy, #782 looks fully redundant rather than complementary — but whether to close it, or instead strip this branch back and land yours first, is his decision, not something to settle automatically. Nothing further from me on it.

Thanks for the cross-check — the `actions/*` finding was correct and is what drove the policy change here.

---
_Generated by [Claude Code](https://claude.ai/code)_

leiicamundi (approved):
Second pass on `3a2ec36`, verified rather than re-read.

## Verified independently

| check | result |
|---|---|
| 37 unique pins re-resolved upstream against the tag in their comment | 37/37 exact, 0 mismatches |
| every pin equals what the ref on `main` resolves to today | 36/36 no-op, nothing silently up- or downgraded |
| gate, third-party on a tag (`mikefarah/yq@v4.53.3`) | `Failed`, exit 14 |
| gate, `actions/*` on a tag (`actions/checkout@v7`) | `Failed` |
| hook coverage | 55 files, 45 of them composite `action.yml` |

The second gate test is the one that matters: that case was green before and is red now. The finding is closed, not worked around.

## The `camunda/*` correction is yours, and it is the better one

You went to the docs where I had gone to the error message, and the wording settles it:

> This includes actions from your organization and actions authored by GitHub. Reusable workflows can still be referenced by tag.

Organization ownership buys no exemption; the carve-out is for reusable workflows, and every `camunda/*` reference here is a step-level composite action. My `camunda/*: hash-pin` suggestion was right by accident and for the wrong reason. Recording the trade-off as a `KNOWN GAP` in `zizmor.yml`, with the three self-references named, is better than either silently exempting them or breaking the `@main` contract inside a pinning PR.

Checked the eight remaining `camunda/*@main` references against our consumers: none of them reach `setup-yarn-cache`, `fossa/info` or `configure-maintenance-pull-request`, so that gap does not block us. Flagging it only because it will block whoever does reach them.

## Counts

Reconciled, and your split was the right one — mine counted the two example `uses:` lines in `README.md` and `docs/github-actions-pinning.md`, which are documentation, not references GitHub resolves. Current state on the branch: 79 pinned occurrences, 37 unique refs, 34 files.

## Still open, not blocking

`.github/workflows/pre-commit.yml` L27: `pre-commit/action` is pinned, but `action.yml` at `2c7b3805` still carries `uses: actions/cache@v4`. Since the setting walks the whole tree, enabling it on this repository refuses the workflow that enforces the policy — the gate cannot run to prove itself. Agreed it is out of scope here and correctly recorded as a follow-up.

## Verdict

Approving. Pins correct, no behavioural change, the gate demonstrably catches both classes it now claims to cover, and the one remaining gap is named rather than hidden.

Closing my #782, it is a strict subset of this branch — my 30 `actions/*` pin lines are byte-identical to yours at the same file and line, which is a decent independent control on both.

PR #789, opened 2026-08-17
fix(actionlint): install from verified release assets instead of curl | bash

### LINKED ISSUE #778: actionlint action: installer is curl|bash from a mutable ref, with no checksum

leiicamundi:
Closes #778.

## The line

```bash
bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash) $ACTIONLINT_VERSION
```

Three things were unverified in it, and this is a shared action, so all three reached every repository that calls it:

1. **The script came from a mutable ref.** `.../actionlint/main/scripts/...` is whatever is on upstream's default branch at the moment the job runs.
2. **It was piped into `bash`** with no `--fail`, so an HTTP error page was executed as a shell script too.
3. **The script it ran verified nothing** — `curl -L "${url}" | tar xvz` with no checksum.

## What replaces it

The release asset for the runner's own OS and architecture, checked against the `SHA256` sums published with that same release, with `--fail --proto '=https' --tlsv1.2 --retry 3`. Nothing is executed to obtain the binary, and the tarball is unpacked in a temp directory rather than the workspace.

Verified by running the step body as it stands in the action:

| case | result |
|---|---|
| linux/amd64 | installs 1.7.10, lints clean |
| linux/arm64 | installs 1.7.10, lints clean |
| darwin/arm64 | installs 1.7.10, lints clean |
| release does not exist | `curl: (22) ... 404`, step fails — previously the 404 body was executed and the job continued |
| tarball altered after download | `checksum mismatch`, expected and actual printed, step fails |

The last one was produced by appending a byte to the downloaded asset before verification, not by argument.

## The `eval`, which turned out to be more than a smell

```bash
_args+=(-ignore "'$line'")
...
eval "./actionlint ${_args[@]}"
```

An apostrophe in an ignore pattern terminated the quoting:

```
eval: line 25: unexpected EOF while looking for matching `''
```

and a pattern shaped like `x' $(command) '` ran the command substitution — I confirmed that by watching a file appear. In practice `ignore` is a literal in the calling workflow, and whoever writes that could add a `run:` step instead, so this is not privilege escalation on its own. It matters because it turns a data input of a shared action into shell syntax, and a caller feeding it anything computed inherits that.

The array is now passed straight through. Behaviour is unchanged for real patterns — including the backtick example from the README, checked side by side against the old form on identical input:

| ignore pattern | old | new |
|---|---|---|
| none | identical | identical |
| `is potentially untrusted` | identical, suppresses | identical, suppresses |
| ``object type "{}" ... `.*` ...`` | identical | identical |
| `it's not a real pattern` | **shell syntax error** | treated as a regex |

## Smaller items

- `REPOSITORY_NAME=${{ github.repository }}` expanded a template into a shell body. `$GITHUB_REPOSITORY` already holds it.
- `version` gains the repository's usual `# renovate: datasource=github-releases depName=rhysd/actionlint` marker. It was the only tool default here without one, which is why it has been bumped by hand (`4ca4aa3`, `bd035ce`).
- `actions/checkout` in this action is already SHA-pinned, so that part of the issue is stale.
- The README documented `1.7.9` while the action defaulted to `1.7.10`.

## Tests

`test-actionlint.yml`, following the shape of `test-common-tooling.yml`:

- the action runs on `ubuntu-latest` and `aws-arm-core-2-default`, because the asset is resolved from `uname` and only a real run on each proves both branches;
- it is called with `it's not a pattern that matches anything`, which is the regression test for the `eval` removal;
- a second job asks for version `0.0.0` and asserts the step **fails**, which is the regression test for `--fail`.

## Not addressed here

The token-permissions passthrough I raised in the issue thread is a different action (`generate-github-app-token-from-vault-secrets`) and belongs in its own PR.

### DIFF
--- a/.github/workflows/test-actionlint.yml
+++ b/.github/workflows/test-actionlint.yml
@@ -1,0 +1,61 @@
+---
+name: Test actionlint
+
+on:
+  pull_request:
+    paths:
+    - 'actionlint/**'
+    - '.github/workflows/test-actionlint.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  # Exercises the install on both architectures the org runs on: the action
+  # resolves the release asset from `uname`, so amd64 and arm64 take different
+  # branches and only a real run on each proves them.
+  test-actionlint:
+    strategy:
+      fail-fast: false
+      matrix:
+        label: ["ubuntu-latest", "aws-arm-core-2-default"]
+    runs-on: ${{ matrix.label }}
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+    - name: Lint this repository with the action
+      uses: ./actionlint
+      with:
+        # The apostrophe is the point. This value used to be pasted into a
+        # command string that was then eval'd, where it terminated the quoting
+        # and produced a shell syntax error; it now has to survive as a plain
+        # regular expression that happens to match nothing.
+        ignore: |-
+          it's not a pattern that matches anything
+
+  # The installer used to pipe whatever the download URL returned into bash, so
+  # an HTTP error page was executed and the job carried on. A release that does
+  # not exist must now stop the step.
+  test-actionlint-rejects-unknown-version:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+    - name: Ask for a release that does not exist
+      id: missing-version
+      continue-on-error: true
+      uses: ./actionlint
+      with:
+        version: "0.0.0"
+
+    - name: The failed download must have failed the step
+      env:
+        OUTCOME: ${{ steps.missing-version.outcome }}
+      run: |
+        if [[ "${OUTCOME}" != "failure" ]]; then
+          echo "expected the action to fail for a non-existent release, got: ${OUTCOME}" >&2
+          exit 1
+        fi
+      shell: bash
--- a/actionlint/action.yml
+++ b/actionlint/action.yml
@@ -56,11 +56,17 @@ runs:
         checksums="actionlint_${ACTIONLINT_VERSION}_checksums.txt"
         release="https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}"
 
-        for file in "${asset}" "${checksums}"; do
-          curl --fail --silent --show-error --location \
-               --proto '=https' --tlsv1.2 --retry 3 \
-               --output "${workdir}/${file}" "${release}/${file}"
-        done
+        # --proto-redir matters as much as --proto: the release URL redirects to
+        # a CDN, and curl allows plain HTTP on redirect by default. Both files
+        # travel this path, so anyone able to downgrade it would serve the
+        # tarball and the checksums that match it.
+        curl --fail --silent --show-error --location \
+             --proto '=https' --proto-redir '=https' --retry 3 \
+             --output "${workdir}/${asset}" "${release}/${asset}"
+
+        curl --fail --silent --show-error --location \
+             --proto '=https' --proto-redir '=https' --retry 3 \
+             --output "${workdir}/${checksums}" "${release}/${checksums}"
 
         expected="$(awk -v a="${asset}" '$2 == a {print $1}' "${workdir}/${checksums}")"
         if [[ -z "${expected}" ]]; then
--- a/actionlint/action.yml
+++ b/actionlint/action.yml
@@ -6,6 +6,7 @@ description: Linting for GitHub workflows
 inputs:
   version:
     description: Actionlint version
+    # renovate: datasource=github-releases depName=rhysd/actionlint
     default: "1.7.10"
   ignore:
     description: Multiline; Ignore errors by messaging using regular expression.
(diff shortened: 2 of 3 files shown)


leiicamundi:
Three of the four findings say this workflow is invalid YAML — that `paths:` and `steps:` entries are not indented under their key, so the jobs will not run. That is not the case, and it is directly checkable.

A block sequence may sit at the same indentation as its parent key; YAML does not require the extra level. The file parses:

```
$ python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/test-actionlint.yml')); ..."
jobs: ['test-actionlint', 'test-actionlint-rejects-unknown-version']
paths: ['actionlint/**', '.github/workflows/test-actionlint.yml']
  test-actionlint: 2 steps
  test-actionlint-rejects-unknown-version: 3 steps
```

It is also the style `test-common-tooling.yml` already uses in this repository, which is where the layout came from.

The decisive part is that the jobs ran on the commit under review, and the review itself was triggered from that same run:

| check | result |
|---|---|
| `test-actionlint (ubuntu-latest)` | SUCCESS |
| `test-actionlint (aws-arm-core-2-default)` | SUCCESS |
| `test-actionlint-rejects-unknown-version` | SUCCESS |
| `pre-commit` (runs actionlint over this repo) | SUCCESS |

A workflow that failed to load could not have produced three green jobs. No change made for these three.

---

The `echo` / `read` finding is a different matter and two thirds of it holds. Measured on the actual parsing block:

| ignore pattern | before | after |
|---|---|---|
| `-n` | **dropped entirely** | `<-n>` |
| `-e` | **dropped entirely** | `<-e>` |
| `  leading spaces matter` | `<leading spaces matter>` | `<  leading spaces matter>` |
| `a\db` | `<a\db>` | `<a\db>` |

`echo` consumes an argument that is exactly `-n` or `-e`, so such a line vanished from the arguments silently. `read` without `IFS=` strips leading whitespace a regex may depend on. Both are now fixed with `printf '%s\n'` and `IFS= read -r` in 31071d7.

The backslash half does not hold: bash's `echo` without `-e` does not interpret escapes, and both forms yield `a\db` unchanged.

This was pre-existing rather than introduced here, but it sits in the same input path this PR is hardening, so it goes in.

.github/workflows/test-actionlint.yml
```diff
@@ -0,0 +1,61 @@
+---
+name: Test actionlint
+
+on:
+  pull_request:
+    paths:
+    - 'actionlint/**'
+    - '.github/workflows/test-actionlint.yml'
```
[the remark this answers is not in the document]
leiicamundi: This does not hold: a block sequence may sit at the same indentation as its parent key. YAML does not require the extra level, and it is the style `test-common-tooling.yml` already uses in this repository, which is where the layout came from.

The file parses:

```
$ python3 -c "import yaml; d = yaml.safe_load(open('.github/workflows/test-actionlint.yml')); ..."
jobs: ['test-actionlint', 'test-actionlint-rejects-unknown-version']
paths: ['actionlint/**', '.github/workflows/test-actionlint.yml']
  test-actionlint: 2 steps
  test-actionlint-rejects-unknown-version: 3 steps
```

And the decisive part: the jobs ran on the commit under review, in the same run that produced this review — `test-actionlint (ubuntu-latest)`, `test-actionlint (aws-arm-core-2-default)` and `test-actionlint-rejects-unknown-version` all green. A workflow that failed to load could not have produced three passing jobs. No change made.

.github/workflows/test-actionlint.yml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Lint this repository with the action
+      uses: ./actionlint
+      with:
+        # The apostrophe is the point. This value used to be pasted into a
+        # command string that was then eval'd, where it terminated the quoting
+        # and produced a shell syntax error; it now has to survive as a plain
+        # regular expression that happens to match nothing.
+        ignore: |-
+          it's not a pattern that matches anything
+
```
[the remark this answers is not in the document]
leiicamundi: This does not hold: a block sequence may sit at the same indentation as its parent key. YAML does not require the extra level, and it is the style `test-common-tooling.yml` already uses in this repository, which is where the layout came from.

The file parses:

```
$ python3 -c "import yaml; d = yaml.safe_load(open('.github/workflows/test-actionlint.yml')); ..."
jobs: ['test-actionlint', 'test-actionlint-rejects-unknown-version']
paths: ['actionlint/**', '.github/workflows/test-actionlint.yml']
  test-actionlint: 2 steps
  test-actionlint-rejects-unknown-version: 3 steps
```

And the decisive part: the jobs ran on the commit under review, in the same run that produced this review — `test-actionlint (ubuntu-latest)`, `test-actionlint (aws-arm-core-2-default)` and `test-actionlint-rejects-unknown-version` all green. A workflow that failed to load could not have produced three passing jobs. No change made.

actionlint/action.yml
```diff
@@ -68,14 +118,16 @@ runs:
             if [[ -z "$line" ]]; then
               continue
             fi
-            _args+=(-ignore "'$line'")
+            _args+=(-ignore "$line")
           done < <(echo "$non_empty_lines")
         fi
```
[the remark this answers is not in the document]
leiicamundi: Two thirds of this holds, and it is fixed in 31071d7. Measured on the parsing block rather than argued:

| ignore pattern | before | after |
|---|---|---|
| `-n` | **dropped entirely** | `<-n>` |
| `-e` | **dropped entirely** | `<-e>` |
| `  leading spaces matter` | `<leading spaces matter>` | `<  leading spaces matter>` |
| `a\db` | `<a\db>` | `<a\db>` |

`echo` consumes an argument that is exactly `-n` or `-e`, so such a line vanished from the arguments with nothing said about it. `read` without `IFS=` strips leading whitespace that a regular expression can depend on. Both now use `printf '%s\n'` and `IFS= read -r`.

The backslash half does not hold: bash's `echo` without `-e` does not interpret escapes, and both forms produce `a\db` unchanged.

This was pre-existing rather than introduced by this branch, but it sits in the same input path the branch is hardening, and silently discarding a caller's pattern is the same failure shape as the rest of it.

actionlint/action.yml
```diff
@@ -30,15 +31,64 @@ runs:
...
+        esac
+
+        workdir="$(mktemp -d)"
+        trap 'rm -rf "${workdir}"' EXIT
+
+        asset="actionlint_${ACTIONLINT_VERSION}_${os}_${arch}.tar.gz"
+        checksums="actionlint_${ACTIONLINT_VERSION}_checksums.txt"
+        release="https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}"
+
+        for file in "${asset}" "${checksums}"; do
```
maxdanilov: (minor)
why do we need a for loop wrapper here, if `asset` and `checksums` are both plain strings (not an array)?

actionlint/action.yml
```diff
@@ -30,15 +31,64 @@ runs:
...
+          curl --fail --silent --show-error --location \
+               --proto '=https' --tlsv1.2 --retry 3 \
```
maxdanilov: why we're hardcoding to TLS 1.2? it's not the most recent one https://endoflife.date/tls anyway

maxdanilov: do we even need the `--tls` flag here?

actionlint/action.yml
leiicamundi: I agree, I think the llm wanted "combine it", now the download part is 2  curl commands with the same flags

actionlint/action.yml
leiicamundi: We don't. --tlsv1.2 is a floor, not a target. The curl negociated tls 1.3 even with the flag, removed for clairty

leiicamundi:
Thanks for flagging the 2 points @maxdanilov. It's now addressed, happy to get a counter review

maxdanilov (approved):
Looks good, thank you!

